PR #115, opened 2025-03-22
fix: clear input fields when closing create task dialog

stoilovska:
**Description:**
When creating a new task, if a user filled in the fields but closed the dialog without submitting, the form data would remain populated the next time the dialog was opened.

This PR fixes the issue by clearing the input fields when the create task dialog is closed.

Now, when you reopen the dialog, all fields are reset to their default empty state.

**Before:** Form retains old values when reopening the dialog.
image

**After:** Form is cleared when reopening the dialog.
image


aacevski (approved):
Thank you so much for fixing the bug. Approved and merged 🙏

PR #133, opened 2025-03-27
feat: disable create project and workspace buttons

stoilovska:
**For Project Creation:**

- When creating a project, if any of the input fields are left empty, the "Create Project" button should be disabled.

- The button should only be enabled when all required fields are filled in.

image

**For Workspace Creation:**

- Similarly, when creating a workspace, if any of the input fields are left empty, the "Create Workspace" button should be disabled.

- The button should only be enabled when all required fields are filled in.

image


aacevski (approved):
Thank you!

PR #138, opened 2025-03-29
feat: migrate to node.js

aacevski:
The goal of this pull request is to migrate to Node.js.

The reason is that Synology runs on Linux kernel version 4.4 and Bun starts support from 5.1. 

On top of that, I'm migrating to Hono as I see their e2e types are much better. 

I'm also updating from Bun -> pnpm workspaces.


apps/api/src/user/controllers/sign-in.ts
```diff
@@ -1,12 +1,14 @@
+import { HTTPException } from "hono/http-exception";
```
aacevski, line 1: No need for these changes here.

apps/hono-api/src/workspace/controllers/get-workspaces.ts
```diff
@@ -0,0 +1,45 @@
...
+    )
+    .where(
+      or(
+        eq(workspaceTable.ownerEmail, userEmail),
+        eq(workspaceUserTable.userEmail, userEmail),
+      ),
+    )
+    .groupBy(workspaceTable.id, workspaceTable.name, workspaceTable.ownerEmail);
+
+  const workspacesWithProjects = await Promise.all(
```
aacevski, line 29: I don't think I need this anymore

apps/hono-api/src/workspace/controllers/update-workspace.ts
```diff
@@ -0,0 +1,51 @@
...
+    })
+    .where(eq(workspaceTable.id, workspaceId))
+    .returning({
+      id: workspaceTable.id,
+      name: workspaceTable.name,
+      ownerEmail: workspaceTable.ownerEmail,
+      createdAt: workspaceTable.createdAt,
+    });
+
+  return updatedWorkspace.at(0);
```
aacevski, line 48:
```suggestion
  const [updatedWorkspace] = await db
    .update(workspaceTable)
    .set({
      name,
      description,
    })
    .where(eq(workspaceTable.id, workspaceId))
    .returning({
      id: workspaceTable.id,
      name: workspaceTable.name,
      ownerEmail: workspaceTable.ownerEmail,
      createdAt: workspaceTable.createdAt,
    });

  return updatedWorkspace;
```

.github/workflows/node-tag.yml
```diff
@@ -0,0 +1,51 @@
+name: Tag Docker images with :node tag
```
aacevski, line 1: Remove before merging.

Issue #122, opened 2025-03-26
Delete task


alexgutjahr:
Would love to be able to delete a task.

aacevski:
Hey! I agree, I'll add this this weekend since it's heavily requested. Thank you for the feedback!

moohbr:
Howdy! I've opened a [PR](https://github.com/usekaneo/kaneo/pull/216), that I **expect** to close the request. Stay tuned for updates!

aacevski:
Closing this as @moohbr's PR has been merged!

PR #216, opened 2025-05-03
feat: delete a task option #122

moohbr:
# Delete task buttton
This PR adds a button that allows users to delete a task, addressing issue #122 . 

## Summary

- Implemented a backend API endpoint to handle task deletion.
- Integrated the deletion logic with the frontend.
- Added a "Delete Task" button in the UI.
- After deletion, users are redirected to the board page.

## Image
image

## Caution
This feature enables task deletion. Since the project currently lacks a role or permission system, any user who can view a task is also able to delete it. This may need to be addressed in future iterations.


aacevski (approved):
Thank you for these changes! It looks great.

PR #1166, opened 2026-04-06
feat(task): track task author and surface as sticker

RainyPixel:
## Summary

In shared workspaces it is currently impossible to tell who created a task — the only user shown on a card is the assignee. This PR adds an explicit `created_by` field to the task table and surfaces it in the UI.

### Backend
- New `created_by` column on `task` (FK → `user`, `ON DELETE SET NULL`) with index, plus a Drizzle migration (`0028_fantastic_thor_girl.sql`).
- `taskTableRelations` gets a new `creator` relation; `userTableRelations` gets `createdTasks`. Both `assignee` and `creator` use explicit `relationName`s so Drizzle can disambiguate the two foreign keys to `user`.
- `createTask` accepts an optional `createdBy` and the `POST /:projectId` route fills it from `c.get("userId")`. Existing tasks remain `null` (the column is nullable, no backfill).
- `getTask` and `getTasks` add a self-join on `user` aliased as `creator_user` and expose `createdBy`, `creatorName`, `creatorImage` in the response.
- `taskSchema` (Valibot, used in OpenAPI) gets `createdBy: nullable(string)`.
- `update-task`, `move-task`, and `bulk-update-tasks` are intentionally left untouched — `createdBy` is immutable like `createdAt`.

### Frontend
- `Task` type gets optional `createdBy`, `creatorName`, `creatorImage` fields.
- **Kanban card** (`task-card.tsx`): when the author differs from the assignee, a compact badge with the author's avatar and a localized "Author" label is rendered next to the priority badge. Hovering shows a tooltip with the author's name.
- **Task properties sidebar** (`task-properties-sidebar.tsx`): a read-only author chip is added next to the assignee chip in all three layouts (compact, mobile, desktop).

### i18n
- New keys: `tasks:creator.label`, `tasks:creator.unknown`, `tasks:creator.tooltip` (with `{{name}}` interpolation).
- Translations added for all nine supported locales (en-US, ru-RU, uk-UA, de-DE, es-ES, fr-FR, nl-NL, mk-MK, el-GR).
- `i18n/schema.json` updated and `creator` added to the `tasks` namespace's `required` list.

## Test plan
- [ ] `pnpm --filter @kaneo/api db:migrate` runs the new migration cleanly on an existing database
- [ ] Creating a new task via the UI sets `created_by` to the current user (verify in DB or via `GET /:id`)
- [ ] Existing tasks created before this change still load (with `createdBy: null`) and the kanban card renders without the author badge
- [ ] Assigning a task to someone else makes the author badge appear on the card and in the sidebar
- [ ] Self-assigned tasks (author = assignee) hide the author badge on the card to avoid duplication
- [ ] When the original creator has been removed from the workspace, the badge falls back to `creatorName`/`creatorImage` returned from the API rather than disappearing
- [ ] Switching the UI language renders the badge label and tooltip in each of the nine supported locales

## New Features

* Tasks now display creator information prominently on kanban board cards and in the task properties sidebar
* Creator details including name and avatar are displayed alongside other task metadata such as assignee and priority
* Full multi-language support for task creator information across all supported locales and languages

### DIFF
--- a/apps/api/drizzle/0028_fantastic_thor_girl.sql
+++ b/apps/api/drizzle/0028_fantastic_thor_girl.sql
@@ -1,3 +1,3 @@
 ALTER TABLE "task" ADD COLUMN "created_by" text;--> statement-breakpoint
-ALTER TABLE "task" ADD CONSTRAINT "task_created_by_user_id_fk" FOREIGN KEY ("created_by") REFERENCES "public"."user"("id") ON DELETE set null ON UPDATE cascade;--> statement-breakpoint
+ALTER TABLE "task" ADD CONSTRAINT "task_created_by_user_id_fk" FOREIGN KEY ("created_by") REFERENCES "public"."user"("id") ON DELETE cascade ON UPDATE cascade;--> statement-breakpoint
 CREATE INDEX "task_createdBy_idx" ON "task" USING btree ("created_by");
--- a/apps/api/drizzle/0028_fantastic_thor_girl.sql
+++ b/apps/api/drizzle/0028_fantastic_thor_girl.sql
@@ -1,0 +1,3 @@
+ALTER TABLE "task" ADD COLUMN "created_by" text;--> statement-breakpoint
+ALTER TABLE "task" ADD CONSTRAINT "task_created_by_user_id_fk" FOREIGN KEY ("created_by") REFERENCES "public"."user"("id") ON DELETE set null ON UPDATE cascade;--> statement-breakpoint
+CREATE INDEX "task_createdBy_idx" ON "task" USING btree ("created_by");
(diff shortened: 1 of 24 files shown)


andrejsshell (changes requested):
When I create a task it says: Andrej Task 4 seconds ago
Image

Also, maybe in the task details sidebar, to add some indicator that the person is the creator, it's confusing if I have the assignee and creator the same: 

Image

RainyPixel:
@andrejsshell thanks for the review! I've addressed all the feedback:

1. **FK onDelete rule** — changed `onDelete: "set null"` to `onDelete: "cascade"` in both schema.ts and SQL migration (follows project convention)

2. **Missing newline** — added trailing newline at end of `0028_fantastic_thor_girl.sql`

3. **Sidebar UX** — added \"Author:\" label before the creator name in task properties sidebar so it's clear who the creator is even when creator = assignee

PR #1203, opened 2026-04-15
fix(chart): add KANEO_CLIENT_URL to web container and api.extraEnv support

awbait:
## Summary

Two independent Helm chart fixes encountered while self-hosting Kaneo v2.6.8 with a custom OIDC provider (Authentik):

- **Add \`KANEO_CLIENT_URL\` to the web container env.** The web image's \`env.sh\` substitutes \`KANEO_CLIENT_URL\` placeholders in the built JS at container start. Currently the chart only sets \`KANEO_API_URL\` on the web container, so \`VITE_CLIENT_URL\` never gets substituted and the compiled bundle keeps the literal string \`KANEO_CLIENT_URL\`. Better Auth then rejects the OAuth sign-in request:
  \`\`\`
  POST /api/auth/sign-in/oauth2
  { \"providerId\": \"custom\", \"callbackURL\": \"KANEO_CLIENT_URL/dashboard\", \"errorCallbackURL\": \"KANEO_CLIENT_URL/auth/sign-in\" }
  -> 403 { \"message\": \"Invalid callbackURL\", \"code\": \"INVALID_CALLBACK_URL\" }
  \`\`\`
  Fix: emit \`KANEO_CLIENT_URL\` alongside \`KANEO_API_URL\`, using the same \`web.env.apiUrl\` fallback logic already used for the API container.

- **Add \`api.extraEnv\`** so custom env vars (\`CUSTOM_OAUTH_*\`, \`SMTP_*\`, \`DEVICE_AUTH_CLIENT_IDS\`, …) can be injected without forking the chart. Standard \`EnvVar\` list, supports both \`value\` and \`valueFrom\`.

Chart bumped 0.2.0 → 0.3.0.

## Test plan
- [x] \`helm lint charts/kaneo\`
- [x] \`helm template\` with \`--set api.extraEnv[0].name=FOO --set api.extraEnv[0].value=bar\` — renders \`FOO\` on api container, \`KANEO_CLIENT_URL\` on web container
- [x] Deployed to my homelab cluster with Authentik as OIDC provider — \`/api/auth/sign-in/oauth2\` now returns 200 and the flow completes successfully (hard-refresh required after upgrade because the JS is cached)

Happy to split into two PRs if preferred.

### DIFF
--- a/charts/kaneo/Chart.yaml
+++ b/charts/kaneo/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: kaneo
 description: Kaneo - All you need. Nothing you don't. Open source project management that works for you.
 type: application
-# Bumped to 0.2.0 due to breaking auth and probe changes
-version: 0.2.0
+# 0.3.0: add api.extraEnv for custom env vars; fix missing KANEO_CLIENT_URL on web container
+version: 0.3.0
 # Bumped to the latest v2 release
 appVersion: "2.3.5"
 keywords:
--- a/charts/kaneo/templates/deployment.yaml
+++ b/charts/kaneo/templates/deployment.yaml
@@ -90,6 +90,9 @@ spec:
               value: {{ .Values.api.env.disableRegistration | quote }}
             - name: DISABLE_PASSWORD_REGISTRATION
               value: {{ .Values.api.env.disablePasswordRegistration | quote }}
+            {{- with .Values.api.extraEnv }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
           livenessProbe:
             {{- toYaml .Values.api.livenessProbe | nindent 12 }}
           readinessProbe:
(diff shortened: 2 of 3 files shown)


tinsever:
Thank you for your contribution! Will look into it tomorrow! :)

andrejsshell:
Thanks for the careful write-up and the homelab validation, @awbait — really appreciate the detail.

Closing this one because the chart was restructured on `main` after you opened it: the API and web containers were collapsed into a single `kaneo` container, and `KANEO_CLIENT_URL` is now set directly on it from `.Values.kaneo.env.clientUrl`. So the OAuth callback bug you hit doesn't exist on the current chart — your fix solved it for the old two-container layout, but the equivalent path now flows through `kaneo.env.clientUrl` instead.

The `extraEnv` escape hatch is still a useful pattern, but with the single-container layout it'd be `kaneo.extraEnv` rather than `api.extraEnv`. We'll track that as a follow-up.

Sorry for the wasted cycles on the rebase — truly grateful for the report and the fix; it would have been the right one if main hadn't moved underneath it.

andrejsshell:
Quick follow-up — to keep credit where it's due, I ported the `extraEnv` half of your PR onto `main` (adapted to the new single-container `kaneo.extraEnv` shape) and committed it with you as the author: ef63690f.

So GitHub still credits you in the contributors graph for the part of this work that survived the chart restructure. Thanks again, @awbait.

andrejsshell:
Tiny correction — fixed the author email so GitHub actually associates the commit with your account. New SHA: 1cfa2e92.

PR #1361, opened 2026-07-02
fix(i18n): make workspace roles settings translatable

bramvera:
## Summary
- route the workspace roles settings page through i18n instead of rendering hardcoded English
- add English reference strings for workspace role headings, defaults, permission labels/descriptions, validation, toasts, and delete confirmation text
- keep fallback text for unknown future dynamic permissions

## Why
The workspace roles settings UI currently renders much of its copy directly in the route component. That makes strings such as role headings, permission labels, status text, and button labels stay in English even when another locale is selected.

## Validation
- `pnpm exec biome ci apps/web/src/routes/_layout/_authenticated/dashboard/settings/workspace/roles.tsx i18n/en-US.json`
- `git diff --check`
- `pnpm --filter @kaneo/permissions build`
- `pnpm --filter @kaneo/email build`
- `pnpm --filter @kaneo/web build`
- `pnpm build`
- `pnpm exec biome ci .`

## Notes
- `pnpm test` currently fails locally in an existing web test, `src/hooks/use-task-filters-with-labels-support.test.tsx`, because `window.localStorage.clear` is not available in the local Vitest environment. This file is unrelated to the diff.
- `pnpm i18n:report` still fails on the current repo baseline with existing missing/unused keys and dynamic-key reporting. The workspace roles permission keys in this PR are accessed with analyzer-friendly dynamic prefixes and were not reported as newly unused.

## Summary by CodeRabbit

* **New Features**
  * Workspace role management text is now fully localized across the settings dashboard.
  * Role names, descriptions, permissions, buttons, dialogs, and status messages now adapt to the selected language.

* **Bug Fixes**
  * Improved consistency for role-related notices, including loading, empty, error, and validation messages.
  * Updated permission and role-count text to support correct pluralization and localized display.

PR #1362, opened 2026-07-02
feat(i18n): add Indonesian locale

bramvera:
## Summary

- Add `id-ID` Indonesian translations.
- Register `id-ID` in the shared i18n resources so it appears in the account language picker.
- Regenerate the i18n JSON schema.

## Dependency

This branch is stacked on top of #1361 so the Indonesian locale includes the new workspace roles strings instead of adding a locale that still falls back to English on the Roles page. Until #1361 is merged, this PR's diff will include that dependency commit too.

## Validation

- `pnpm i18n:check id-ID` passes.
- `pnpm exec biome ci i18n/id-ID.json i18n/resources.ts i18n/schema.json` passes.
- `pnpm --filter @kaneo/web build` passes.
- Runtime tested on a self-hosted Kaneo 2.9.0 deployment using this branch plus #1361: health endpoint OK, sign-in page renders without React error #321, `id-ID` workspace role strings are present in the deployed bundle, and the Roles route chunk is present.

Known repo-wide i18n status:

- `pnpm i18n:check` still fails because existing locales are missing keys, including the new workspace roles keys from #1361.
- `pnpm i18n:report` still reports pre-existing missing/unused/dynamic-key findings.

## Summary by CodeRabbit

* **New Features**
  * Added Indonesian language support across the app.
  * Localized the workspace roles settings experience, including titles, labels, actions, and status messages.

* **Bug Fixes**
  * Improved consistency of translated text across role management, permissions, notifications, tasks, and settings screens.
  * Added missing wording for several empty, loading, error, and confirmation states.

* **Chores**
  * Expanded language coverage and updated translation structure to support additional localized content.

### DIFF
--- a/apps/web/src/routes/_layout/_authenticated/dashboard/settings/workspace/roles.tsx
+++ b/apps/web/src/routes/_layout/_authenticated/dashboard/settings/workspace/roles.tsx
@@ -2,6 +2,7 @@ import { DEFAULT_ROLE_NAMES, statement } from "@kaneo/permissions";
 import { createFileRoute } from "@tanstack/react-router";
 import { Plus, Shield, Trash2, X } from "lucide-react";
 import { useMemo, useState } from "react";
+import { useTranslation } from "react-i18next";
 import PageTitle from "@/components/page-title";
 import {
   Accordion,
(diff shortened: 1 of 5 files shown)


tinsever:
Thank you a lot!

PR #1404, opened 2026-07-17
fix(web): load status options from columns query

### LINKED ISSUE #1402: [Bug]: Cannot change ticket status in full-screen view
### Bug Description

When opening a ticket in full-screen mode, it is impossible to change the ticket status from the properties sidebar. Clicking the status button triggers a UI issue and prevents selecting a new status.

Important detail: This bug ONLY occurs after a page refresh or direct URL load. If you navigate to the ticket through the standard UI and expand it to full-screen without refreshing the page, the status dropdown works perfectly.

### Steps To Reproduce

1. Open a ticket in full-screen view.
2. Refresh the browser tab (or load the full-screen ticket URL directly).
3. Go to the "Properties" sidebar on the right and try to change the status.

### Expected Behavior

The status dropdown should open normally in full-screen view, just as it does in the regular/modal ticket view, allowing the user to select a different status.

### Actual result

The status dropdown is completely unresponsive and cannot be changed. Other property fields work perfectly fine.

### Screenshots

Image

Image

### Environment

Browser: Arc Browser (Latest)
OS: macOS 26 Tahoe
Kaneo Version: 2.9.0 Self-Hosted
Device: MacBook Air M3

### Additional Context

_No response_

andrejsshell:
## Summary
The status dropdown in the task properties sidebar rendered zero options on a direct load or refresh of the full-screen task view, making the status impossible to change. TaskStatusPopover read columns from the Zustand project store, which only the board, backlog, and list routes populate; the full-screen route never does.

## Changes
- TaskStatusPopover fetches columns with the useGetColumns query (keyed by task.projectId) instead of reading the project store, matching how SubtaskStatusPopover already works
- Remove the now-unused project store import

## Test Plan
- Root cause confirmed by code inspection: only the board, backlog, and list routes call setProject, so the store is empty on a direct full-screen load; SubtaskStatusPopover uses the query and does not exhibit the bug
- 19 web unit tests pass, Biome clean, full monorepo build passes
- Not manually reproduced in a browser; worth a quick manual check of the repro steps from the issue before merging

Fixes #1402

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task status selection by loading available statuses more reliably.
  * Preserved status labels, icons, completion indicators, and keyboard shortcut behavior in the status menu.

### DIFF
--- a/apps/web/src/components/task/task-status-popover.tsx
+++ b/apps/web/src/components/task/task-status-popover.tsx
@@ -9,12 +9,12 @@ import {
 } from "@/components/ui/popover";
 import { ShortcutNumber } from "@/components/ui/shortcut-number";
 import { useUpdateTaskStatus } from "@/hooks/mutations/task/use-update-task-status";
+import { useGetColumns } from "@/hooks/queries/column/use-get-columns";
 import { useNumberedShortcuts } from "@/hooks/use-numbered-shortcuts";
 import { useWorkspacePermission } from "@/hooks/use-workspace-permission";
 import { getColumnIcon } from "@/lib/column";
 import { getStatusDisplayLabel } from "@/lib/i18n/domain";
 import { toast } from "@/lib/toast";
-import useProjectStore from "@/store/project";
 import type Task from "@/types/task";
 
 type TaskStatusPopoverProps = {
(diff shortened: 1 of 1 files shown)


andrejsshell:
Addressed the review-bot findings in a0a4edef:

- **CodeRabbit (memoize statusOptions):** applied. Options are now built in a useMemo keyed on the query data, so the shortcutOptions memo holds between renders.
- **qodo (loading/error UX for the popover):** skipped. The properties sidebar fetches the same query key for every viewer, so the data is cached before the popover opens in practice, and the option list fills in reactively if a fetch is still in flight. Dedicated loading/error states here would be more UI than the popover warrants.
- **qodo (columns fetched before permission check):** skipped. Rules of Hooks require the hook to run unconditionally, and the surrounding views already fetch the same query key for read-only users, so gating the fetch would save no network traffic.

PR #1408, opened 2026-07-18
feat: webhook events, unicode slugs, board sort

### LINKED ISSUE #1357: feat: Persist sort/filter selection across page reloads on the Kanban board
### Problem Statement

Currently, any sort or filter applied to the Kanban board (e.g. sorting by priority) resets as soon as the page is reloaded. It would be great if the selected sort/filter state persisted across reloads, so users don't have to reapply it every time.

### Proposed Solution

Store the current sort/filter state (e.g. in local storage or as a URL query parameter) so it's restored automatically on page load.

This would improve the workflow for users who regularly work with a specific board view.

### Alternative Solutions

Instead of full persistence, a simpler alternative could be to at least remember the sort/filter within the same session (e.g. via a state manager or session storage), even if it resets on a full logout or browser restart. Another option would be to let users save a specific view as a "default" for a project, rather than persisting the last-used state automatically.

### Relevant Context

This becomes noticeable when regularly working on a board with a specific sort (e.g. priority) or filter applied — every page reload (navigation, refresh, etc.) forces the user to reapply it manually, which adds friction to daily use of the Kanban board.

Yes, because it doesn't add any new UI or complexity — it simply preserves an existing user choice (sort/filter) instead of silently discarding it. It removes a repetitive manual step rather than adding one, which is in line with keeping the tool simple and frictionless to use.

### LINKED ISSUE #1395: [Bug]: Cannot create column with non-Latin name (Cyrillic, Arabic, Chinese, etc.)
### Bug Description

Column names containing only non-Latin characters (Cyrillic, Arabic, Chinese, etc.)
cannot be created, while renaming an existing column to the same name works fine.

## Root Cause

`toSlug()` in `apps/api/src/column/controllers/create-column.ts` strips all 
non-`[a-z0-9]` characters, producing an empty slug for any non-Latin name,
which triggers the validation error.

```ts
function toSlug(name: string): string {
  return name
    .toLowerCase()
    .trim()
    .replace(/[^a-z0-9]+/g, "-")  // non-Latin → "-"
    .replace(/^-+|-+$/g, "");     // "-" → ""  ← empty slug
}
```

## Suggested Fix

Add the [`transliteration`](https://www.npmjs.com/package/transliteration) package 
(MIT, zero dependencies, supports all Unicode scripts) and update `toSlug()` in 
`apps/api/src/column/controllers/create-column.ts`:

```ts
import { slugify } from "transliteration";

function toSlug(name: string): string {
  const slug = slugify(name, {
    lowercase: true,
    separator: "-",
    trim: true,
  });

  if (!slug && name.trim()) {
    return `col-${Math.random().toString(36).slice(2, 8)}`;
  }

  return slug;
}
```

This produces readable, deterministic slugs for any script
("тест" → "test", "测试" → "ce-shi") while keeping a random
fallback for edge cases where transliteration yields no output.

### Steps To Reproduce

1. Open any project → Settings → Columns
2. Type a non-Latin name in the new column input (e.g. "тест")
3. Click "Add"

### Expected Behavior

Column is created successfully.

### Actual result

Error is shown: `Column name must contain at least one alphanumeric character`

### Screenshots

_No response_

### Environment

- Kaneo Version: 2.9.0
- Deployment Method: Docker
- Browser: Ungoogled Chromium 150.0.7871.46 (arm64)
- OS: macOS

### Additional Context

_No response_

### LINKED ISSUE #1396: feat: Add missing task events to Generic Webhook plugin
### Problem Statement

Several task events exist internally in Kaneo but are never forwarded to Generic Webhook (or any other plugin integration). This makes it impossible to build complete external notification pipelines.

Currently missing from webhook:
| Event | Currently handled by |
| ----- | ----------------------|
| task.deleted | WebSocket + Activity log |
| task.moved | WebSocket + Activity log |
| task.due_date_changed | WebSocket + Activity log |
| task.assignee_changed | WebSocket only |
| task.unassigned | WebSocket only |

`task.assignee_changed` and `task.unassigned` are not even recorded in the activity log - assignee changes are completely invisible outside the UI.

### Proposed Solution

Add handler functions in [apps/api/src/plugins/generic-webhook/events.ts](https://github.com/usekaneo/kaneo/blob/main/apps/api/src/plugins/generic-webhook/events.ts) for each missing event, following the same pattern as existing handlers, with appropriate `data` payloads.

### Alternative Solutions

Expose a catch-all webhook event that forwards all internal events, letting the consumer filter what they need.

### Relevant Context

We are using Generic Webhook to forward Kaneo task events to Mattermost. Assignee changes and due date updates are critical for team coordination but currently require manually checking Kaneo.

Yes. It does not introduce new concepts, only extends the existing webhook plugin to cover events that already exist internally.

andrejsshell:
## Summary
Fixes three open issues in one pass: columns can now be created with non-Latin names (#1395), the Kanban board's sort selection survives page reloads (#1357), and the Generic Webhook plugin forwards the five task events that previously never left the app (#1396). Each issue is a separate commit for easier review.

## Changes
- **Non-Latin column names (`fix(api)`)**: `toSlug()` now uses NFKC normalization with a `\p{L}\p{M}\p{N}` keep-class, so Cyrillic/Arabic/CJK/Indic names produce valid slugs while existing Latin slugs stay byte-identical. Combining marks stay attached (Turkish dotted İ, Devanagari matras).
- **Board sort persistence (`feat(web)`)**: new `use-board-sort` hook mirrors the existing per-project filter persistence (`kaneo:board-sort:<projectId>`), with persisted values validated against `Record`-derived maps so a future `SortField` change fails at compile time instead of silently resetting saved sorts.
- **Webhook task events (`feat`)**: `task.deleted`, `task.moved`, `task.due_date_changed`, `task.assignee_changed`, and `task.unassigned` now flow publisher → registry → generic-webhook, each behind a new opt-in toggle (settings UI, fetchers, OpenAPI schema, all 12 locales). `task.deleted` builds its envelope from event data via a shared delivery/health helper since the row is already gone. Bulk operations now emit the same webhook data as single edits (old assignee/due date, title) and publish `task.deleted` only after the delete succeeds.

Known scope limits, left for follow-up discussion: `task.moved` notifies only the destination project's webhooks (the WS layer notifies both sides), and bulk operations fan out one background webhook POST per task.

## Test Plan
- New unit tests: `toSlug` (7 cases incl. Turkish/Devanagari), `use-board-sort` (restore/fallback/write-through), generic-webhook handlers (disabled paths + enabled `task.deleted` envelope + missing-project guard)
- Full suites green: API 200/200, web 34/34, `biome ci` clean, monorepo build passes (pre-commit hook ran both on every commit)
- Multi-agent review over the full diff (10 finder angles, per-finding verification); all confirmed findings fixed or documented above

Closes #1395
Closes #1357
Closes #1396

https://claude.ai/code/session_01UkZ21HKnpGqwN7p2tkr6CV

## Summary by CodeRabbit

* **New Features**
  * Generic webhooks now support task deletion, movement, due-date changes, assignee changes, and unassignment events.
  * Added new opt-in toggles for these events in generic webhook integration settings, including localized labels.
* **Improvements**
  * Board sorting is now saved/restored per project.
  * Slug generation now properly normalizes Unicode, including full-width text, and avoids empty slugs for non-alphanumeric input.
* **Tests**
  * Added/expanded tests for webhook delivery, new settings behavior, board-sort persistence, and Unicode slugification.

### DIFF
--- a/apps/api/src/column/controllers/create-column.ts
+++ b/apps/api/src/column/controllers/create-column.ts
@@ -5,12 +5,14 @@ import { columnTable } from "../../database/schema";
 import { VIRTUAL_STATUSES } from "../../task/validate-task-fields";
 
 export function toSlug(name: string): string {
-  return name
+  const slug = name
     .normalize("NFKC")
     .toLowerCase()
     .trim()
     .replace(/[^\p{L}\p{M}\p{N}]+/gu, "-")
     .replace(/^-+|-+$/g, "");
+
+  return /[\p{L}\p{N}]/u.test(slug) ? slug : "";
 }
 
 async function createColumn({
--- a/apps/api/src/column/controllers/create-column.ts
+++ b/apps/api/src/column/controllers/create-column.ts
@@ -4,11 +4,12 @@ import db from "../../database";
 import { columnTable } from "../../database/schema";
 import { VIRTUAL_STATUSES } from "../../task/validate-task-fields";
 
-function toSlug(name: string): string {
+export function toSlug(name: string): string {
   return name
+    .normalize("NFKC")
     .toLowerCase()
     .trim()
-    .replace(/[^a-z0-9]+/g, "-")
+    .replace(/[^\p{L}\p{M}\p{N}]+/gu, "-")
     .replace(/^-+|-+$/g, "");
 }
 
--- a/apps/api/src/generic-webhook-integration/index.ts
+++ b/apps/api/src/generic-webhook-integration/index.ts
@@ -85,6 +85,11 @@ const genericWebhookEventsSchema = v.object({
   taskTitleChanged: v.optional(v.boolean()),
   taskDescriptionChanged: v.optional(v.boolean()),
   taskCommentCreated: v.optional(v.boolean()),
+  taskDeleted: v.optional(v.boolean()),
+  taskMoved: v.optional(v.boolean()),
+  taskDueDateChanged: v.optional(v.boolean()),
+  taskAssigneeChanged: v.optional(v.boolean()),
+  taskUnassigned: v.optional(v.boolean()),
   dueDateReminder: v.optional(v.boolean()),
 });
 
--- a/apps/api/src/plugins/generic-webhook/config.ts
+++ b/apps/api/src/plugins/generic-webhook/config.ts
@@ -95,6 +95,11 @@ export const genericWebhookEventKeys = [
   "taskTitleChanged",
   "taskDescriptionChanged",
   "taskCommentCreated",
+  "taskDeleted",
+  "taskMoved",
+  "taskDueDateChanged",
+  "taskAssigneeChanged",
+  "taskUnassigned",
   "dueDateReminder",
 ] as const;
 
@@ -134,6 +139,11 @@ export const genericWebhookConfigSchema = v.object({
       taskTitleChanged: v.optional(v.boolean()),
       taskDescriptionChanged: v.optional(v.boolean()),
       taskCommentCreated: v.optional(v.boolean()),
+      taskDeleted: v.optional(v.boolean()),
+      taskMoved: v.optional(v.boolean()),
+      taskDueDateChanged: v.optional(v.boolean()),
+      taskAssigneeChanged: v.optional(v.boolean()),
+      taskUnassigned: v.optional(v.boolean()),
       dueDateReminder: v.optional(v.boolean()),
     }),
   ),
@@ -156,6 +166,11 @@ export const defaultGenericWebhookEvents: Record<
   taskTitleChanged: false,
   taskDescriptionChanged: false,
   taskCommentCreated: true,
+  taskDeleted: false,
+  taskMoved: false,
+  taskDueDateChanged: false,
+  taskAssigneeChanged: false,
+  taskUnassigned: false,
   dueDateReminder: false,
 };
 
--- a/apps/api/src/plugins/generic-webhook/events.ts
+++ b/apps/api/src/plugins/generic-webhook/events.ts
@@ -10,12 +10,17 @@ import {
 } from "../../database/schema";
 import type {
   PluginContext,
+  TaskAssigneeChangedEvent,
   TaskCommentCreatedEvent,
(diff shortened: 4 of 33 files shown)


andrejsshell:
Addressed the bot review feedback:

**Fixed**
- **Mark-only slugs bypassing validation** (CodeRabbit + Qodo) — `toSlug()` now returns an empty slug unless the result contains at least one `\p{L}` or `\p{N}`, so names made only of combining marks are rejected with the existing 400 while marks attached to real letters are preserved. Regression tests added. (e3d5cb2e)
- **Unguarded `localStorage.setItem`** (Qodo) — the write effect in `use-board-sort` now treats persistence as best-effort with a try/catch, so private-mode/quota failures can't surface as uncaught errors. (aedf045a) Note: `use-task-filters-with-labels-support` has the same unguarded write; leaving that for a follow-up since it predates this PR.

**Declined, with rationale**
- **Qodo: fallback slug for symbol-only names like `"!!!"`** — rejecting these with a clear 400 is pre-existing, intended behavior; #1395's actual bug was non-Latin *rejection*, which is fixed. A random fallback would trade a deterministic, explained error for invisible nondeterministic slugs.
- **CodeRabbit: lazy-init rewrite of `use-board-sort` ("visible flash", mount write race)** — the flash isn't reachable: the board renders a skeleton until project data loads (project store is empty on reload and populated in an effect), so the first non-skeleton paint already has the restored sort. The mount-time double write is real but self-heals within one effect cascade and matches the shipped filter-hook pattern; diverging from that pattern here would leave two side-by-side hooks with different persistence semantics. Happy to revisit both hooks together in a follow-up if wanted.
- **CodeRabbit: #1357 only partially addressed (sort but not filters)** — filter persistence already exists (`kaneo:board-filters:<projectId>`, added in af096952); this PR fills the remaining gap, so the issue is fully satisfied.
- **Enabled-path tests for the other four handlers** — they all route through the same `sendEvent` → shared delivery helper that the `task.deleted` test exercises; the added coverage per mock complexity is low. Can add if maintainers want it.

PR #1434, opened 2026-07-26
fix(email): stop forcing SMTP auth when no credentials are set

### LINKED ISSUE #1419: [Bug]: SMTP authentication is forced for unauthenticated relays
### Bug Description

Kaneo cannot send workspace invitation emails through an SMTP relay that does not require authentication.

The same SMTP relay works correctly with Nodemailer when no `auth` object is configured. However, Kaneo always initializes the Nodemailer transport with:

```ts
auth: {
  user: process.env.SMTP_USER,
  pass: process.env.SMTP_PASSWORD,
}
```

When these variables are not configured, Nodemailer still attempts SMTP authentication.

### Steps To Reproduce

1. Configure Kaneo with an SMTP relay that does not require authentication.
2. Set `SMTP_HOST`, `SMTP_PORT`, `SMTP_SECURE`, and `SMTP_FROM`.
3. Do not set `SMTP_USER` or `SMTP_PASSWORD`.
4. Start Kaneo.
5. Create a workspace or open an existing workspace.
6. Invite another user by email.

### Expected Behavior

Kaneo should send the workspace invitation email without attempting SMTP authentication when `SMTP_USER` and `SMTP_PASSWORD` are not configured.

The Nodemailer transport should omit the `auth` property entirely in this case.

### Actual result

The invitation email is not sent.

Kaneo logs:

```text
Error sending workspace invitation email
Error: Missing credentials for "PLAIN"
```

When SMTP credentials are added, Kaneo instead attempts authentication and the relay rejects it:

```text
Error: Invalid login: 535 authentication failed
...
code: 'EAUTH'
command: 'AUTH PLAIN'
```

### Screenshots

*No response*

### Environment

- Kaneo Version: 2.9.8

### Additional Context

The current implementation appears to always pass an `auth` object to Nodemailer:

```ts
const transporter = nodemailer.createTransport({
  host: process.env.SMTP_HOST,
  secure: process.env.SMTP_SECURE !== "false",
  port: Number(process.env.SMTP_PORT),
  auth: {
    user: process.env.SMTP_USER,
    pass: process.env.SMTP_PASSWORD,
  },
  requireTLS: process.env.SMTP_REQUIRE_TLS === "true",
  ignoreTLS: process.env.SMTP_IGNORE_TLS === "true",
});
```

The `auth` property should only be included when both `SMTP_USER` and `SMTP_PASSWORD` are defined.

For example:

```ts
const smtpUser = process.env.SMTP_USER;
const smtpPassword = process.env.SMTP_PASSWORD;

const transporter = nodemailer.createTransport({
  host: process.env.SMTP_HOST,
  secure: process.env.SMTP_SECURE !== "false",
  port: Number(process.env.SMTP_PORT),
  ...(smtpUser && smtpPassword
    ? {
        auth: {
          user: smtpUser,
          pass: smtpPassword,
        },
      }
    : {}),
  requireTLS: process.env.SMTP_REQUIRE_TLS === "true",
  ignoreTLS: process.env.SMTP_IGNORE_TLS === "true",
});
```

This would support both authenticated SMTP servers and unauthenticated internal relays.

---
<sub>Task: zwstnmddfmhpvotmsc9yrhbs</sub>

eeshsaxena:
Closes #1419.

`packages/email/src/send-email.tsx` builds the nodemailer transport with an `auth` object that is always present:

```ts
auth: { user: process.env.SMTP_USER, pass: process.env.SMTP_PASSWORD }
```

Nodemailer reads the presence of `auth` as "authenticate this session", so with `SMTP_USER` and `SMTP_PASSWORD` unset it still tries to authenticate, and a relay that accepts unauthenticated mail refuses. Workspace invitations never arrive.

Transport options now come from `getSmtpTransportOptions()`, which adds `auth` only when both a user and a password are set. It also leaves `port` out when `SMTP_PORT` is unset instead of passing `Number(undefined)`, which was NaN, so nodemailer can fall back to the default port for the chosen security mode.

The same assumption sits in a second place. `hasSmtp` in `apps/api/src/utils/get-settings.ts` required `SMTP_USER`, `SMTP_PASSWORD`, `SMTP_PORT` and `SMTP_SECURE`, so the unauthenticated relay case also made the API report SMTP as unconfigured and the sign-in page hide the email option. It never checked `SMTP_FROM`, which the send helpers do require, so the two definitions of "SMTP is set up" had drifted apart. Both now come from one `isSmtpConfigured()` (host plus sender), which is what `sendWorkspaceInvitationEmail` and `sendNotificationEmail` already guard on.

`packages/email/src/smtp-config.test.ts` covers no credentials, one credential, both credentials, the missing port, the TLS flags, and both `isSmtpConfigured` outcomes.

One caveat on my end: npm cannot reach the registry from this machine, so I could not install the workspace and run vitest. I ran the same assertions against the real `smtp-config.ts` using node's TypeScript stripping instead, 15 of them, all passing. The vitest file mirrors those cases, so CI should be the real check.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SMTP configuration detection for email sending.
  * Prevented email attempts when required SMTP settings are incomplete.
  * Improved handling of optional SMTP ports, credentials, and TLS settings.
  * Ensured email-related settings accurately reflect whether SMTP is available.

* **Tests**
  * Added coverage for SMTP configuration and transport options across supported settings.
  * Added validation for incomplete credentials and missing required settings.

### DIFF
--- a/apps/api/src/utils/get-settings.ts
+++ b/apps/api/src/utils/get-settings.ts
@@ -1,3 +1,4 @@
+import { isSmtpConfigured } from "@kaneo/email";
 import { config } from "dotenv-mono";
 import { isGithubSsoConfigured } from "./github-sso-env";
 
@@ -10,12 +11,7 @@ function getSettings() {
       process.env.DISABLE_PASSWORD_REGISTRATION === "true",
     disableEmailOtpSignIn: process.env.DISABLE_EMAIL_OTP_SIGN_IN === "true",
     isDemoMode: process.env.DEMO_MODE === "true",
-    hasSmtp:
-      Boolean(process.env.SMTP_HOST) &&
-      Boolean(process.env.SMTP_PORT) &&
-      Boolean(process.env.SMTP_SECURE) &&
-      Boolean(process.env.SMTP_USER) &&
-      Boolean(process.env.SMTP_PASSWORD),
+    hasSmtp: isSmtpConfigured(),
     hasGithubSignIn: isGithubSsoConfigured(),
     hasGoogleSignIn:
       Boolean(process.env.GOOGLE_CLIENT_ID) &&
--- a/packages/email/src/index.tsx
+++ b/packages/email/src/index.tsx
@@ -5,3 +5,4 @@ export {
   sendPasswordResetEmail,
   sendWorkspaceInvitationEmail,
 } from "./send-email";
+export { isSmtpConfigured } from "./smtp-config";
--- a/packages/email/src/send-email.tsx
+++ b/packages/email/src/send-email.tsx
@@ -1,6 +1,7 @@
 import { render } from "@react-email/components";
 import { config } from "dotenv-mono";
 import * as nodemailer from "nodemailer";
+import { getSmtpTransportOptions, isSmtpConfigured } from "./smtp-config";
 import type { MagicLinkEmailProps } from "./templates/magic-link";
 import MagicLinkEmail from "./templates/magic-link";
 import NotificationEmail, {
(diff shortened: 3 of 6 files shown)


randoneering:
@eeshsaxena

Thank you for contribution. Please take a look at the results from the two AI reviews. Additionally (and could be related), CI is failing during API Integration tests

```bash

 FAIL  ../../tests/api-integration/config.test.ts > API integration: config > returns the public config shape
AssertionError: expected 500 to be 200 // Object.is equality

- Expected
+ Received

- 200
+ 500

 ❯ ../../tests/api-integration/config.test.ts:10:29
      8|     const response = await app.request("/api/config");
      9|
     10|     expect(response.status).toBe(200);
       |                             ^
     11|     const payload = (await response.json()) as Record<string, unknown>;
     12|


```

PR #1468, opened 2026-08-02
fix: prevent task number gaps during partial import failures

Sunil56224972:
## Problem

`importTasks()` pre-claimed ALL task numbers upfront via `claimTaskNumbers(projectId, tasksToImport.length)` before inserting tasks one-by-one. If any individual task insert failed (caught in the try/catch), the claimed task number for that slot was permanently wasted - creating gaps in the task numbering sequence.

**Example:** Importing 10 tasks where 3 fail means task numbers jump from N to N+10, but only 7 tasks exist. Numbers at the failed positions are gone forever.

## Root Cause

`claimTaskNumbers()` was called once before the loop, pre-allocating a contiguous block. Each failed insert consumed a number that could never be reclaimed since the claim had already committed.

## Fix

Wrap each task insert in its own `db.transaction()` with `claimTaskNumber()` called **inside** the transaction. If the insert fails, the transaction rolls back and the number is never claimed.

## Changes

- `apps/api/src/task/controllers/import-tasks.ts`: Moved from batch `claimTaskNumbers()` to per-task `claimTaskNumber()` inside individual transactions.

## Testing

- Verified that successful imports maintain contiguous numbering
- Verified that failed imports do not consume task numbers

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task-number assignment during task imports.
  * Ensured each imported task receives a unique number as part of its database transaction, reducing the risk of numbering conflicts.

### DIFF
--- a/apps/api/src/task/controllers/import-tasks.ts
+++ b/apps/api/src/task/controllers/import-tasks.ts
@@ -8,7 +8,7 @@ import {
   coerceStatus,
   getValidTaskStatuses,
 } from "../validate-task-fields";
-import claimTaskNumbers from "./claim-task-numbers";
+import { claimTaskNumber } from "./claim-task-numbers";
 
 export type ImportTask = {
   title: string;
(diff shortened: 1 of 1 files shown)


randoneering:
@Sunil56224972  same here!

PR #1493, opened 2026-08-05
fix(mcp): share OAuth state across replicas

### LINKED ISSUE #1484: [Bug]: Remote MCP OAuth returns invalid_client when requests hit different Kaneo replicas
### Bug Description

The built-in remote MCP OAuth flow fails when Kaneo runs with multiple replicas.

Dynamically registered OAuth clients, pending authorization requests and authorization codes are stored in process-local `Map` instances.

If `/api/mcp/register` reaches replica A and `/api/mcp/authorize` reaches replica B, replica B cannot find the registered client and returns:

```json
{"error":"invalid_client"}
```

The issue was reproduced deterministically with Kaneo v2.12.1. Sending both requests to the same replica succeeds, while sending them to different replicas produces the error.

### Steps To Reproduce

1. Deploy Kaneo v2.12.1 with two replicas:
  ```yaml
   replicaCount: 2
  ```
2. Select two different Kaneo pods and forward them to separate local ports:
  ```bash
   kubectl -n <namespace> port-forward pod/<pod-a> 5173:5173
  ```
3. Register an OAuth client on pod A:
  ```bash
   curl -sS \
     -H 'Content-Type: application/json' \
     --data '{
       "client_name": "replica-test",
       "redirect_uris": ["http://127.0.0.1:43210/callback"],
       "grant_types": ["authorization_code"],
       "response_types": ["code"],
       "token_endpoint_auth_method": "none"
     }' \
     http://127.0.0.1:5173/api/mcp/register
  ```
4. Copy the returned `client_id`:
  ```bash
   CLIENT_ID="<returned-client-id>"
  ```
5. Send the authorization request to pod A:
  ```bash
   curl -i --get http://127.0.0.1:5173/api/mcp/authorize \
     --data-urlencode response_type=code \
     --data-urlencode client_id="$CLIENT_ID" \
     --data-urlencode redirect_uri=http://127.0.0.1:43210/callback \
     --data-urlencode code_challenge=E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM \
     --data-urlencode code_challenge_method=S256
  ```

   This returns a `302` consent redirect.
6. Send the same request to pod B by changing the URL port to `5174`.
7. Observe:
  ```json
   {"error":"invalid_client"}
  ```

### Expected Behavior

The MCP OAuth flow should work independently of which Kaneo replica handles each request.

Registered clients, authorization requests and authorization codes should be available to every replica for their required lifetime.

### Actual result

Registration succeeds on one replica, but another replica does not recognize the returned `client_id`:

```text
pod A: POST /api/mcp/register   → 200
pod A: GET /api/mcp/authorize  → 302
pod B: GET /api/mcp/authorize  → 400 {"error":"invalid_client"}
```

Running a single replica avoids the error but removes high availability.

### Screenshots

*No response*

### Environment

- Kaneo Version: v2.12.1
- Helm Chart Version: 2.12.1
- Deployment Method: Kubernetes using the official Helm chart
- Kaneo Replicas: 2
- Service Session Affinity: None

### Additional Context

OAuth clients, authorization requests and authorization codes are stored in process-local maps:

[https://github.com/usekaneo/kaneo/blob/v2.12.1/apps/api/src/mcp/oauth.ts#L29-L31](https://github.com/usekaneo/kaneo/blob/v2.12.1/apps/api/src/mcp/oauth.ts#L29-L31)

The authorization endpoint returns `invalid_client` when the current process cannot find the client:

[https://github.com/usekaneo/kaneo/blob/v2.12.1/apps/api/src/mcp/controllers/oauth-consent.ts#L76-L80](https://github.com/usekaneo/kaneo/blob/v2.12.1/apps/api/src/mcp/controllers/oauth-consent.ts#L76-L80)

Authorization-code exchange also depends on process-local state:

[https://github.com/usekaneo/kaneo/blob/v2.12.1/apps/api/src/mcp/oauth.ts#L124-L137](https://github.com/usekaneo/kaneo/blob/v2.12.1/apps/api/src/mcp/oauth.ts#L124-L137)

A durable fix would require shared storage for OAuth clients, authorization requests and codes, with expiration and atomic consume semantics.

An integration test should verify that registration, authorization and token exchange can be handled by different Kaneo instances.

---
<sub>Task: vlsaemg6dv14g3mpg55tflz7</sub>

Oxygen56:
## Description

- Persist MCP OAuth clients, authorization requests, and authorization codes in PostgreSQL so every Kaneo replica sees the same state.
- Atomically consume authorization requests and codes to preserve one-time-use behavior across replicas.
- Add expiry indexes, cleanup, a generated migration, and a cross-module-instance integration test.

## Related Issue(s)

Fixes #1484

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Test addition or update

## How Has This Been Tested?

- [x] Unit tests
- [x] Integration tests
- [ ] Manual testing
- [x] Other: type checking and Biome checks on the changed files

Commands run:

- `pnpm --pm-on-fail=ignore --filter @kaneo/api exec vitest run --config vitest.config.ts` (253 tests)
- `DATABASE_URL=postgresql://oxygen@localhost:5432/kaneo_test pnpm --pm-on-fail=ignore --filter @kaneo/api exec vitest run --config vitest.integration.config.ts` (100 tests)
- `pnpm --pm-on-fail=ignore typecheck`
- `pnpm --pm-on-fail=ignore exec biome ci <changed files>`
- `pnpm --pm-on-fail=ignore --filter @kaneo/api db:generate` (no schema drift)

Not applicable.

## Checklist

- [x] New and existing API unit tests pass locally with my changes

## Additional Notes

The integration test resets the module graph between OAuth steps to exercise the same persistence boundary used by separate application replicas. Authorization requests and codes remain single-use through `DELETE ... RETURNING` operations.

## Summary by CodeRabbit

- **New Features**
  - MCP OAuth state is now persisted, enabling authorization data to be shared across application instances.
  - Added support for durable client registrations, authorization requests, and authorization codes.
  - Authorization codes and requests now expire and are safely consumed only once.

- **Bug Fixes**
  - Improved redirect URI and PKCE validation during authorization and code exchange.

- **Tests**
  - Added integration and security coverage for shared state, expiration, single-use codes, and validation.


eeshsaxena (commented):
Read this focused on the code-exchange path since that's the security-critical part, and the state move to Postgres looks sound for the replica-sharing goal:

- `exchangeCode` consumes the code with `DELETE ... RETURNING`, so single-use is enforced atomically across replicas: a second exchange of the same code finds no row and returns null. That's exactly the guarantee the in-memory `Map` couldn't give once there is more than one replica.
- Deleting the code before validating `clientId` / `redirectUri` / expiry / PKCE means a failed exchange still consumes the code, which lines up with RFC 6749's guidance to revoke an authorization code when a bad exchange is attempted.
- `getAuthorizationRequest` and `consumeAuthorizationRequest` both null-check the destructured row before dereferencing, and expiry is enforced in the `WHERE` (`gt(expiresAt, now())`), so a missing or expired request returns undefined instead of throwing.

One thing worth an explicit guard: `verifyPkce` always hashes with SHA-256, so it assumes `code_challenge_method: S256`. If a client ever registers a `plain` challenge, a valid verifier would be rejected at exchange time. If only S256 is supported (typical for MCP), rejecting a non-S256 method at registration/authorization would make that fail clearly rather than surfacing as a confusing "invalid code" later.

andrejsshell:
Hey @Oxygen56, thank you for this PR, and apologies for the awkward timing: I was implementing the same fix for #1484 independently on the same day you opened this, and my version landed on main (18ad8f53) and shipped in v2.13.0 before I'd seen your work.

I went through your implementation carefully rather than just closing this, and two parts of it were better than what we shipped, so they're now on main with your credit:

- **Your cross-replica integration test** is merged under your authorship in 353b38d2. It exercises the exact #1484 scenario against a real database, register on one "replica", consume the consent request on a second, exchange and burn the code on a third with a real S256 verifier — coverage our own tests didn't have.
- **Your PR caught that the Postgres move silently dropped the 10,000-row bound** on pending authorization requests that the old in-memory store enforced. That's restored in 3da2ac5f, credited to you as co-author.

The reason for keeping the merged implementation was structural: it stores all three OAuth kinds in one generic table with a single sweep, and gives dynamically registered clients a 30-day TTL so the unauthenticated register endpoint can't grow the table forever.

Closing as superseded. Since the test lands under your authorship, you'll appear in the contributors list and the release notes. Thanks again. This was a genuinely useful contribution, and I'd be glad to see more from you here.

PR #1519, opened 2026-08-08
fix(task): preserve labels on project import

eeshsaxena:
## Problem

Exporting a project and importing it back silently loses every task's labels.

`export-tasks` serialises labels for each task:

```ts
tasks: tasks.map((task) => ({
  ...
  labels: taskLabelsMap.get(task.id) || [],   // [{ name, color }, ...]
})),
```

But `import-tasks` never reads them: `ImportTask` has no `labels` field, the import route's validator strips the key, and the controller only inserts the task row. So an export -> import round-trip drops all labels.

## Fix

- Add `labels?: Array<{ name: string; color: string }>` to `ImportTask`.
- Accept `labels` in the `/import/:projectId` route validator (otherwise valibot strips it before the controller runs).
- Inside the same transaction that creates the task, re-create its label rows.

Labels are de-duplicated by name before insert, and the insert uses `onConflictDoNothing()` so the `(task_id, name)` unique constraint is respected. `workspaceId` is left null, which is how task-scoped labels are already stored.

The change is additive: exports made before this still import fine (labels default to none), and imports without a `labels` field behave exactly as before.

## Summary by CodeRabbit

* **New Features**
  * Task imports now accept optional labels with names and colors.
  * Imported labels are trimmed, deduplicated, and added to newly created tasks.

* **Bug Fixes**
  * Task deletion now verifies the required entitlement before proceeding.

### DIFF
--- a/apps/api/src/task/controllers/import-tasks.ts
+++ b/apps/api/src/task/controllers/import-tasks.ts
@@ -96,6 +96,7 @@ async function importTasks(
                 name: label.name,
                 color: label.color,
                 taskId: task.id,
+                workspaceId: project.workspaceId,
               })),
             )
             .onConflictDoNothing();
--- a/apps/api/src/task/controllers/import-tasks.ts
+++ b/apps/api/src/task/controllers/import-tasks.ts
@@ -1,7 +1,12 @@
 import { and, eq } from "drizzle-orm";
 import { HTTPException } from "hono/http-exception";
 import db from "../../database";
-import { columnTable, projectTable, taskTable } from "../../database/schema";
+import {
+  columnTable,
+  labelTable,
+  projectTable,
+  taskTable,
+} from "../../database/schema";
 import { publishEvent } from "../../events";
 import {
   coercePriority,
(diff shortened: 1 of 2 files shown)


randoneering:
@eeshsaxena thank you for your contribution! Please take a look at the suggestions from coderabbit and qodo. In addition, our lint step in CI is failing (specifically in Biome.) This should have thrown when you committed to the branch as well.

eeshsaxena:
Thanks for the review.

**coderabbit's suggestion**: added a `trim()` + non-empty filter on label names before the insert, and skip the insert entirely when nothing remains. **qodo's `workspaceId` note** is already handled by this PR — the label insert sets `workspaceId: project.workspaceId`.

**On the Biome/lint failure**: none of the diagnostics are in this PR's two files (`import-tasks.ts`, `task/index.ts`) — they're `useOptionalChain` / `` / empty-import warnings across `apps/web`, `apps/site` and other `apps/api` files. This branch is ~127 commits behind `main`, so `biome ci .` is running against a stale tree. Happy to rebase onto current `main` so the lint runs against your latest state — want me to, or would you rather update the branch on merge?

eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

PR #1520, opened 2026-08-08
fix(web): respect granular task permissions

### LINKED ISSUE #1505: [Bug]: member with 'task-update' permission cannot modify tasks
### Bug Description

Hi,

this is probably a similar issue to https://github.com/usekaneo/kaneo/issues/1345

Members of a project are able to modify the description of a task and can drag&drop them to other lanes. But have no permission to:
* adjust the state ('Todo', 'In progress', 'Blocked' via the dropdown. It _does_ work via drag & drop !)
* adjust the dates
* create/modify labels
* modify priorities

Member permission scheme:
* projects: Read
* task: Create, Read, Update, Assign
* labels: Create, Read, Update
* workspace: Read

Admin users have no issue to perform these modifications on the same tasks. 
When granting an affected user 'admin' privileges, he is instantly able to make the required changes to the date, labels & priorities.

### Steps To Reproduce

Assign the default 'member' permissions to a user:

Member permission scheme:
* projects: Read
* task: Create, Read, Update, Assign
* labels: Create, Read, Update
* workspace: Read

Let the user login. He's able to create a card, but cannot modify dates, labels and priorities of cards. Issue is resolved when the user gains 'admin' privileges.

### Expected Behavior

user objects should be able to perform actions for which permissions were granted. In this case: modify state, dates, labels and priorities of cards.

### Actual result

Nothing happens. Buttons for the date are greyed out, so the user already gets an idea that those aren't available. The state, priority, labels etc.. buttons are not greyed out and also provide no user feedback. From a UX point of view this is not OK, as users gets confused

### Screenshots

Image

### Environment

Kaneo version: v2.9.2
Deployed via Docker compose
No error messages visisble in the logs
No indication of an error in the UI
Happens on all browsers

### Additional Context

Not necessarily related to this issue, but make sure to provide decent feedback if an action is not allowed. Either remove the options in the UI, make them greyed out + tooltip or display a popup/notification when clicked.

Otherwise: keep up the good work 👍

tinsever:
## Description

Respect task and label permissions individually in the task editing UI. Members with task update permission can edit task fields without requiring task delete permission.

## Related Issue(s)

Fixes #1505

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test addition or update

## How Has This Been Tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] Manual testing

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added granular permissions for creating, updating, deleting, assigning, and labeling tasks.
  * Bulk actions now appear individually based on available permissions.
  * Label assignment and removal are more reliable, with separate controls for creating and editing labels.
  * Task fields, subtasks, relations, and descriptions now respect update permissions.
  * Workspace label creation, editing, and deletion are independently controlled.

* **Bug Fixes**
  * Unauthorized users no longer see task editing, attachment, or mutation controls.
  * Subtask creation now requires task creation permission.
  * Users can view labels without seeing unauthorized editing controls.


luantaraschi (commented):
Checked this out at `5edd50dd`, alongside #1559, which lands on the same issue. Since two PRs are open against #1505 I measured how they relate rather than reviewing this one in isolation.

**The core of the fix is byte-identical in both PRs.** Nine of the seventeen files #1559 touches are the same content at both heads, and each is a two-line change in #1559:

```
column-header.tsx, subtask-status-popover.tsx, task-due-date-popover.tsx,
task-priority-popover.tsx, task-relations.tsx, task-start-date-popover.tsx,
task-status-popover.tsx, task-status-popover.test.tsx, task-title.tsx
```

Both split `manageTasks: { task: ["create","update","delete"] }` into `updateTasks` and `deleteTasks`, with the same capability names and the same helper names. So on the task half of #1505 there is no disagreement between the two, and #1559's seventeen files are a strict subset of this PR's twenty-seven.

**The difference is the label half, and #1505 asks for it explicitly.** The report lists four broken things: state, dates, labels, priorities. #1559 stops at `manageLabels`, which still requires `label: ["create","update","delete"]` together:

```ts
// #1559, task-labels-popover.tsx
const { canUpdateTasks, canManageLabels } = useWorkspacePermission();
const canCreateLabels = canManageLabels();
```

The reporter's role grants labels create, read and update, with no delete, which is exactly the combination that fails that check. So #1559 fixes three of the four symptoms for that role and leaves the label one where it was. This PR splits the label capability into create, update and delete, which is what closes the fourth.

**The label gating here matches the server, and the comment it removes did not.** I went looking for a problem in gating attach and detach on `canUpdateLabels`, because the comment on `main` says attaching a label to a task is a task mutation. The API disagrees with that comment. Both routes in `apps/api/src/label/index.ts` guard on the label capability:

```ts
.put("/:id/task", ..., requireWorkspacePermission({ label: ["update"] }), ...)
.delete("/:id/task", ..., requireWorkspacePermission({ label: ["update"] }), ...)
```

So this PR's `canUpdateLabels()` mirrors what the server will actually enforce, and gating on task update would have shown controls that then fail with a 403. The new fetchers also target routes that already exist on `main`, so nothing here waits on an API change.

**What I ran**, on Windows with Node 22.20.0 and pnpm 10.34.1:

| tree | test files | tests |
| --- | --- | --- |
| `main` `3fc77f00` | 27 passed | 82 passed |
| #1520 `5edd50dd` | 26 passed | 75 passed |
| #1559 `a588f776` | 24 passed | 71 passed |

Each tree also has the same three failures, and all three are mine rather than anyone's code: two files import `@kaneo/permissions`, which my install did not build, and `src/env.test.ts` shells out to `sed`, which Windows does not have. The counts differ because the branches sit on different bases, not because tests were removed.

For whoever decides: this PR is the superset and the one that closes #1505 as filed. #1559 is a clean, smaller change that agrees with it line for line on the task half, so if you would rather land the small one first, the two do not conflict in intent, only in size. What I would avoid is merging #1559 and treating #1505 as done, since the label case that the reporter described would still be broken.

Thanks to both of you for working the same bug carefully.

PR #1525, opened 2026-08-08
feat(project): move a project to another workspace

### LINKED ISSUE #1328: feat: relocate project from one workspace to another?
### Problem Statement

We have reorganized our teams and need to reassign projects to different workspaces but can't see how to do this in Kaneo. 

### Proposed Solution

It's common to rethink how projects are assigned or grouped. Is this possible with cloud version? 

### Alternative Solutions

Could use self-hosted and try to find relevant database fields to change but feel this is a pretty core need. 

### Relevant Context

Maybe could add a reassignment field in settings. 

Seems like it's not an addition but just a basic way to insure folks aren't locked into a rigid data org.

rdlugs:
Closes #1328

## Problem

There's no way to reassign a project to a different workspace. Teams that reorganize are stuck either recreating the project by hand or editing the database directly, which is what the issue reporter was resorting to.

## What this adds

`PUT /project/:id/move` plus a **Move to workspace** control in project general settings.

**Permissions.** The caller needs `project:update` *and* `project:delete` in the source workspace — a move removes the project from there — and `project:create` in the target. Every built-in role that grants `update` also grants `delete`, so the pair only matters for custom roles, where `update` alone would otherwise let someone move a project out of a workspace they can't delete from. The target is authorized separately (membership, permission, entitlement) because the access middleware only resolves one workspace per request.

**Key collisions.** The project key doubles as the ticket-id prefix (`KAN-12`), and short-id lookup resolves it per workspace with a limit of 1. If the target workspace already uses the key, the move is refused with a 409 naming the incumbent project, rather than silently making ticket ids ambiguous. Compared case-insensitively, matching the lookup.

**Data that moves with it.** All in one transaction:

- `project.workspaceId` is repointed; tasks, columns, and comments follow by project id.
- `asset` and `label` rows denormalize the workspace, so they're rewritten.
- Project notification subscriptions are deleted — they reference the project through a composite `(workspace_id, project_id)` key, and the cascade from the workspace update would break the rule side, since the rule itself stays behind in the source workspace.
- Tasks assigned to people who aren't members of the target workspace are unassigned, each with its own `task.unassigned` event so the change lands on the task timeline. Events publish after commit, so a rollback can't leave activity rows behind for a move that never happened.

The confirm dialog spells out the unassignment and subscription clearing before the user commits, and the resulting toast reports how many tasks lost their assignee.

## Testing

`tests/api-integration/project-move.test.ts` — 8 integration tests covering: non-member target, target without `project:create`, key collision (including differing case), same key left behind in the source not blocking, source without `project:delete`, moving into the current workspace, the full side-data rewrite, and assignee dropping with its activity row.

Full suite green: `pnpm typecheck`, `pnpm test`, `pnpm test:integration` (117 tests), `pnpm lint`, `pnpm build`.

## Notes

- All 17 locales updated along with `i18n/schema.json`.
- Entitlement is enforced on the target workspace only, matching how `update` and `archive` behave — an expired source workspace can still move projects out.
- Stored S3 object keys keep their original `workspace/<id>/…` prefix after a move. Serving and cleanup both use the stored key verbatim and authorize against the asset row's workspace, so this is cosmetic; rewriting the keys would mean copying objects.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Projects can now be moved between workspaces from project settings.
  * Destination workspaces are filtered by access and permissions, with confirmation and outcome notifications.
  * Tasks assigned to people without destination access are unassigned and reported.
  * Workspace-specific project data and notification settings are updated during the move.
* **Localization**
  * Added project-move translations across supported languages.
* **Tests**
  * Added coverage for permissions, conflicts, successful moves, data updates, and task reassignment.


randoneering:
@rdlugs  thank you for your contribution! Please take a look at the coderabbit suggestions when you have a chance! Thank you!

rdlugs:
Thanks! I take a look at the coderabbit suggestions and address them accordingly, appreciate the feedback

PR #1526, opened 2026-08-08
fix(github): keep a webhook delivery alive when link metadata will not parse

luantaraschi:
Follow-up to the two things I flagged while reviewing #1518 rather than folding them into that diff.

### The metadata parse

`externalLink.metadata` is a JSON string in the database. A row written by an older version, or truncated, is not something the incoming delivery can do anything about, but four handlers on the GitHub side called `JSON.parse` on it bare:

| file | line |
|---|---|
| `webhooks/issue-closed.ts` | 57 |
| `webhooks/issue-edited.ts` | 67 |
| `webhooks/pull-request-closed.ts` | 63 and 85 |

The throw escapes the handler, so the whole webhook delivery fails and the task is left unsynced with nothing in the log explaining why.

The Gitea handlers already do this properly. `gitea/webhooks/issue-closed.ts` and `gitea/webhooks/issue-reopened.ts` parse inside a `try`, warn with the external link id, and carry on with an empty object. This does the same, in one helper, since the GitHub side needs it in four places rather than one.

### The early return

`issue-closed` returned after the first matching integration, so a repository connected to two projects moved one task to Done and left the other behind. Its Gitea twin walks all of them, and so does `issue-opened` on this side, so this looked like the odd one out rather than a decision.

`pull-request-closed` returns early on both sides, so I left that one alone. If the intent there is also to process every integration, that is a separate change and I would rather ask than assume.

### What I deliberately did not touch

`JSON.parse(integration.config)` is still bare in `issue-opened`, `pull-request-opened`, `push` and `pull-request-closed`. The Gitea side guards those too, but config is written by the app rather than arriving from outside, so the failure story is much weaker and it did not feel like it belonged in the same diff. Happy to send it separately if you want the parity.

### Tests

`tests/api/plugins/github/utils/parse-link-metadata.test.ts`, three cases: a well formed row, an absent value, and a truncated one where it warns and returns an empty object instead of throwing.

The webhook handlers themselves have no tests in the repo today, since they reach the database, so the coverage here is on the helper. Worth saying plainly rather than implying more than there is.

`npx vitest run tests/api/plugins/github` gives 7 files and 38 tests passing, and `biome check --error-on-warnings` is clean on the five files touched.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of GitHub issue and pull request webhook metadata when values are missing, malformed, or invalid.
  - Webhook processing now continues across all applicable integrations instead of stopping after the first one.
  - Added contextual warning logs for invalid metadata without interrupting webhook processing.
  - Improved issue-edit synchronization when timestamp metadata is unavailable.

- **Tests**
  - Added coverage for valid, missing, malformed, and non-object metadata scenarios.

### Sequencing with #1518

#1518 adds `webhooks/issue-reopened.ts`, which does this same parse with its own inline `try`, written before this helper existed. The two do not touch the same files so they will not conflict, but if both land the repo ends up with five parses where four go through the helper and one does not.

Whichever merges second, I will send a one line follow-up moving `issue-reopened` onto the helper. Say the word if you would rather I fold it into one of them now instead.


luantaraschi:
The typecheck was right to fail and I had not run it before opening this, only vitest and biome. Fixed in `32a4447`.

`JSON.parse` was returning `any`, so `metadata.lastSync.title` in `issue-edited` was never checked by anything. The helper hands back a typed value, which is what made `tsc` speak up. It takes the shape from the caller now, defaulting to the untyped record for the handlers that only spread the value forward, and the edit handler declares what it reads.

One behavioural note, since it is the kind of thing worth stating rather than hiding in a diff: a stamp with no `timestamp` used to reach `new Date(undefined)`, an Invalid Date whose `NaN` failed the `< 2000` recency comparison. It now reaches `new Date(0)`, and a value that large fails the same comparison, so the branch taken is unchanged either way.

`tsc --noEmit` reports nothing in the files this PR touches. The `@kaneo/email` and `@kaneo/permissions` errors in that output are workspace packages that are not built in my checkout and are present on `main` too.

luantaraschi:
Both bots found the same real gap, and it was in the part of this PR that was supposed to close it. Fixed in `cdf3ce2a`.

`JSON.parse` does not have to answer an object. The row `null` parses to `null` and `"kaneo"` parses to a string, and the helper handed either one back typed as `Partial<T>`. `handleIssueClosed` then reads `existingMetadata.createdFrom` off it and throws inside the webhook, which is the crash this PR exists to remove. The `try/catch` never sees it, because nothing threw during parsing. Arrays go out with them: nothing here writes one, and a handler that only spreads the value forward would turn its indices into keys.

The premise of the helper is a row written by an older version or truncated, so a row holding `null` belongs in the same bucket as a row that will not parse. It takes the same exit and logs.

I also took the point about the log. The warning no longer carries the row: `issue-edited.ts` stores `formattedDescription` in that metadata, so a truncated row can hold a task description synced from Kaneo. The `externalLinkId` is already there and is enough to find the row by, and there is a test that the description does not reach the log.

Checked the tests catch it rather than just pass. Reverting the guard alone leaves the three tests that were already here green and fails both new ones:

```
 ❯ tests/api/plugins/github/utils/parse-link-metadata.test.ts (5 tests | 2 failed)
     × keeps the row out of the warning
     × ignores a row that parses to something other than an object
AssertionError: expected null to deeply equal {}
```

`npx vitest run tests/api/plugins/github` is 40 passing across 7 files, and `biome check` is clean on both files. `tsc --noEmit` reports nothing in what this PR touches; the `@kaneo/email` and `@kaneo/permissions` errors are workspace packages that are not built in my checkout and are on `main` too.

PR #1529, opened 2026-08-09
feat(task): let images be resized in the task description

### LINKED ISSUE #1211: feat(task): resize image task card
## Problem Statement
When adding images to a card, there’s no option to resize them—only show full image when clicked. It would be really nice to have a resize option, especially for vertical images, since they tend to take up too much space.

## Proposed Solution
- Let users choose image size (small, medium, large)
- Allow simple drag-to-resize in the UI

## Alternative Solutions
- Use smaller thumbnails by default for all images
- Add a toggle to show/hide image previews
- Limit the maximum height of images inside cards

## Relevant Context
This issue is most noticeable with vertical images, especially on smaller screens, where they take up too much space

## Does this feature align with Kaneo's focus on simplicity?
Yes, users can focus on essential information without being distracted by oversized visual elements.

shiminshen:
## Description

Images in a task description always render at full size, which crowds the
description on smaller screens — vertical images especially.

This adds a resizable image node view with the two sizing affordances the
issue asks for: presets (small / medium / large, as a fraction of the editor
width) and a drag handle on the image edge. Controls reveal on hover, because
clicking an editor image already opens the preview dialog.

The non-obvious part is persistence. Descriptions are saved as markdown, and
`alt` has nowhere to carry a width — so an image attribute alone is
dropped on the next load, and a resize would appear to work until reload. A
resized image is therefore serialized as an `` tag carrying the width,
the same HTML-in-markdown route `attachment-card` already uses, and parsed
back on load. Unresized images are still written as plain markdown, so
existing descriptions are byte-for-byte untouched.

Widths are pixel counts; a percentage in an inline style is ignored rather
than misread as pixels, and `src`/`alt`/`title` are escaped on the way out
via the existing `escapeHtml` helper.

Prior art: #1367 took a similar approach before being closed by its author.
This is an independent implementation.

## Related Issue(s)

Fixes #1211

## Type of Change

## How Has This Been Tested?

- [x] Unit tests
- [x] Manual testing

`apps/web/src/components/task/extensions/resizable-image.test.ts` covers
8 cases: plain markdown for unresized images, HTML serialization when
resized, the full markdown round-trip restoring the width, HTML escaping,
parsing a width off an attribute and off an inline pixel style, rejecting a
percentage style, and rejecting a non-positive width.

Full web suite: 77 passing (was 69). `pnpm typecheck` and `biome ci .` clean.

## Checklist

- [x] I have added tests that prove my feature works

## Additional Notes

`pnpm i18n:check:fix` added the five new keys to the other locale files with
the English source text, per the workflow in CONTRIBUTING. Happy to drop those
files from the PR if you would rather translations landed separately.

## Summary by CodeRabbit

* **New Features**
  * Added image resizing in the task editor with drag controls, size presets, and reset functionality.
  * Preserved image dimensions when saving and reopening task descriptions.
  * Improved handling of image URLs, alt text, and special characters during formatting.

* **Localization**
  * Added labels and accessibility text for image resizing controls across supported languages.

* **Tests**
  * Added coverage for resizing, serialization, width validation, escaping, and HTML handling.

### DIFF
--- a/apps/web/src/components/task/extensions/resizable-image.test.ts
+++ b/apps/web/src/components/task/extensions/resizable-image.test.ts
@@ -109,4 +109,62 @@ describe("ResizableImage", () => {
 
     expect(instance.getJSON().content?.[0]?.attrs?.width).toBeNull();
   });
+
+  it("ignores a percentage in the width attribute", () => {
+    const instance = createEditor();
+    instance.commands.setContent(`<img src="${SRC}" width="50%">`);
+
+    expect(instance.getJSON().content?.[0]?.attrs?.width).toBeNull();
+  });
+
+  it("accepts an explicit px suffix on the width attribute", () => {
+    const instance = createEditor();
+    instance.commands.setContent(`<img src="${SRC}" width="320px">`);
+
+    expect(instance.getJSON().content?.[0]?.attrs?.width).toBe(320);
+  });
+
+  it("ignores a width attribute in a unit that is not pixels", () => {
+    const instance = createEditor();
+    instance.commands.setContent(`<img src="${SRC}" width="20em">`);
+
+    expect(instance.getJSON().content?.[0]?.attrs?.width).toBeNull();
+  });
+
+  it("falls back to the inline style when the width attribute is unusable", () => {
+    const instance = createEditor();
+    instance.commands.setContent(
+      `<img src="${SRC}" width="50%" style="width: 320px">`,
+    );
+
+    expect(instance.getJSON().content?.[0]?.attrs?.width).toBe(320);
+  });
+
+  it("escapes brackets in the alt text of an unresized image", () => {
+    const instance = createEditor();
+    setImage(instance, { alt: "report].png" });
+
+    const markdown = instance.getMarkdown();
+    instance.commands.setContent(markdown, { contentType: "markdown" });
+
+    const image = instance.getJSON().content?.[0];
+
+    expect(image?.type).toBe("image");
+    expect(image?.attrs?.src).toBe(SRC);
+    expect(image?.attrs?.alt).toBe("report].png");
+  });
+
+  it("keeps a parenthesised src parseable for an unresized image", () => {
+    const instance = createEditor();
+    const parenSrc = "https://kaneo.app/diagram(1).png";
+    instance.commands.setContent({
+      type: "doc",
+      content: [{ type: "image", attrs: { src: parenSrc, alt: "Diagram" } }],
+    });
+
+    const markdown = instance.getMarkdown();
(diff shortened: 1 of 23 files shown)


randoneering:
@shiminshen  thank you for your contribution. Please take a look at the suggestions by coderabbit and qodo. In addition, there are a few merge conflicts. Thanks!

shiminshen:
Thanks for taking a look. Rebased onto `main` — conflicts were confined to the i18n files and are resolved — and worked through the CodeRabbit and Qodo findings. Both tools flagged the same three defects independently, and all six reproduced, so I wrote tests first and fixed against them.

**`parseWidth` accepting trailing garbage** — the important one. `Number.parseInt` stops at the first non-digit, so `width="50%"` became a 50px image, and because `??` in `parseHTML` short-circuits on any non-null attribute the pixel-only guard never saw it. It now matches the whole value and takes a pixel count with an optional `px`, rejecting every other unit. Reading the attribute and the inline style independently also fixes a related case: an unusable `width="50%"` used to shadow a perfectly good `style="width: 320px"`.

**Markdown escaping** — this was worse than the review suggested. The plain-markdown branch escaped nothing while the HTML branch escaped everything, so an uploaded file named `report].png` didn't just produce broken markdown: the `]` closed the alt text early and the image came back as a *paragraph* on the next load, i.e. silently destroyed. Alt text and titles are escaped now, and a src containing whitespace moves into the `<...>` form.

**Resize performance** — `updateAttributes` ran per `pointermove`, so every pointer event triggered a ProseMirror transaction plus the full-document markdown re-serialization in `TaskDescription.onUpdate`. The drag now previews in local state and commits once on release. That also makes a single undo reverse the whole gesture, which I think is better than the `addToHistory: false` approach the review suggested.

**`max-width` desync** — correct, `min(100%, 44rem)` still won above 704px, so a resized image rendered at a size different from the one persisted. An explicit resize now lifts the readability cap.

**Touch and keyboard** — controls are visible where hover doesn't exist, and the handle takes arrow keys (shift for a coarser step) instead of being focusable but inert.

Seven new tests cover the percentage attribute, the `px` suffix, a non-pixel unit, the style fallback, and the bracket and parenthesis round-trips. Full suite is green (126 tests), typecheck clean, `biome ci` clean.

One note on the i18n diff: running `pnpm i18n:check:fix` also wanted to backfill a batch of pre-existing `common:error.*` keys that are missing from the locale files on `main`, and regenerating `i18n/schema.json` pulls in the same keys. I kept both scoped to just the new `image` keys so this PR doesn't carry unrelated churn — happy to open a separate PR for that backfill if it's wanted. (`pnpm i18n:check` already exits 1 on a clean `main` for the same reason.)

PR #1556, opened 2026-08-11
fix: use internal URL for MCP tool requests

### LINKED ISSUE #1547: MCP tools return 'fetch failed' on localhost self-hosting: server-side calls use the public KANEO_API_URL
## Summary

On a self-hosted install reached over `localhost`, every built-in MCP tool returns:

```json
{ "error": "fetch failed" }
```

OAuth completes fine and the server shows as connected — `whoami`, `list_workspaces`, `list_projects` all fail the same way. The cause is that the MCP handler performs **server-side** fetches against the **public** URL, which does not resolve from inside the container.

## Root cause

`apps/api/src/mcp/index.ts` derives its API base from the public env var:

```js
var apiUrl2 = (process.env.KANEO_API_URL || "http://localhost:1337").replace(/\/api\/?$/, "");
```

`kaneo-entrypoint.sh` derives `KANEO_API_URL` from `KANEO_CLIENT_URL` when unset:

```
KANEO_API_URL not set — derived from KANEO_CLIENT_URL: http://localhost:5273/api
```

That value is correct for the browser and wrong for a server-side fetch. Verified in the running container (`/proc/<api-pid>/environ`):

```
KANEO_API_URL=http://localhost:5273/api
KANEO_CLIENT_URL=http://localhost:5273
```

Inside the container that URL is unreachable, for two compounding reasons:

```
$ cat /etc/hosts | grep localhost
::1               localhost  localhost        # no 127.0.0.1 mapping

$ netstat -tln
0.0.0.0:5173    LISTEN   nginx: master        # IPv4 only
:::1337         LISTEN   node

$ wget -qO- http://localhost:5273/api/health  ->  refused   ("fetch failed")
$ wget -qO- http://127.0.0.1:5173/api/health  ->  {"status":"ok"}
$ wget -qO- http://127.0.0.1:1337/api/health  ->  {"status":"ok"}
```

So: `localhost` resolves only to `::1`, nginx listens only on IPv4, and nothing listens on the published host port *inside* the container. The MCP client is asking the container to reach itself at an address that only exists on the host.

With a public domain this is invisible — the container resolves the hostname and comes back in over the network. It breaks specifically for `localhost` / no-domain deployments (cf. #593), and it is the mirror image of #1218: there the browser wrongly got an internal address; here a server-side caller wrongly gets the public one. One variable is serving two consumers with opposite requirements.

## Steps to reproduce

1. Compose install with `KANEO_CLIENT_URL=http://localhost:<port>` and `KANEO_API_URL` unset (as `.env.sample` suggests).
2. `claude mcp add --transport http kaneo http://localhost:<port>/api/mcp`, complete the OAuth flow.
3. Call any MCP tool → `{"error":"fetch failed"}`.

## Environment

- `ghcr.io/usekaneo/kaneo:2.16.0`, Docker Compose, `postgres:16-alpine`
- Docker Engine 29.6.2 / Compose v5.3.1
- Host port 5273 → container 5173, `KANEO_CLIENT_URL=http://localhost:5273`

## Suggested fix

Server-side callers should not use the public URL. The API already knows where it listens, so the MCP client can target loopback directly:

```js
const internalApiUrl =
  process.env.KANEO_INTERNAL_API_URL ||
  `http://127.0.0.1:${process.env.PORT || 1337}`;
```

Two details worth keeping in mind while fixing it:

- **Prefer the `127.0.0.1` literal over `localhost`.** The existing fallback `http://localhost:1337` only works by luck: `localhost` is `::1` in this image and node happens to bind `:::1337` (dual-stack). Any component that binds IPv4-only — nginx already does — is unreachable via `localhost` from inside. The image's own healthcheck already uses `127.0.0.1` for this reason.
- The same reasoning applies to any other server-side fetch built from `KANEO_API_URL` (e.g. the asset URL builder), where the public value is correct for what it returns to clients but not for outbound calls.

## Workaround for other self-hosters

Without patching the image, make the container reachable at the address it advertises — add the published port as an extra nginx listener inside the container, leaving every public URL untouched (so CORS, cookies and the MCP OAuth registration keep working):

```sh
# runs before the normal entrypoint; note BusyBox sed needs `a\`, not `\n` in s///
sed -i "/listen 5173;/a\\    listen [::]:5273;" /etc/nginx/conf.d/default.conf
sed -i "/listen 5173;/a\\    listen 5273;"      /etc/nginx/conf.d/default.conf
```

Wired in compose as:

```yaml
volumes:
  - ./fix-mcp-loopback.sh:/opt/fix-mcp-loopback.sh:ro
command:
  - /bin/sh
  - -c
  - sh /opt/fix-mcp-loopback.sh; exec /usr/local/bin/kaneo-entrypoint.sh
```

(A file dropped into `/docker-entrypoint.d/` is *not* enough — `kaneo-entrypoint.sh` invokes `env.sh` by name rather than iterating the directory.)

After that, all MCP tools work against the same instance with no other change.

---
<sub>Task: zcu9f44iu2h0npj0dtpnp1lh</sub>

alloutflo:
## Description

Separate the public API origin advertised through MCP OAuth metadata from the internal origin used by server-side MCP tool requests. HTTP MCP tools now call `KANEO_INTERNAL_API_URL` when set, otherwise `http://127.0.0.1:1337`. Public OAuth discovery and `WWW-Authenticate` metadata continue to use `KANEO_API_URL`.

## Root Cause

The HTTP MCP handler used one `KANEO_API_URL` value for both public client metadata and server-side fetches. In localhost Docker installations, that public URL points at a host-published port that is not reachable from inside the container, so built-in MCP tools fail with `fetch failed`.

## Impact

Self-hosted localhost deployments can use built-in MCP tools without adding an internal nginx listener or changing their public URL. Existing public-domain and OAuth URLs stay unchanged, and installations with a different internal address can override the documented default.

## Related Issue(s)

Fixes #1547

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [x] Test addition or update

## How Has This Been Tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] Manual testing
- [x] Other: API typecheck, API build, full monorepo build, and Biome CI

Not applicable; this changes server-side MCP request routing only.

## Checklist

## Additional Notes

Validation completed with the repository-pinned pnpm 10.32.1:

- focused MCP tests: 30 passed, including default and configured internal URL routing
- full API unit suite: 321 passed
- API lint and typecheck
- API build and full monorepo build
- `biome ci .` (successful; only pre-existing repository warnings)

No dependent changes are required. The unchecked comment and dependent-change items are not applicable.

## Summary by CodeRabbit

- **New Features**
  - Added support for separate public and internal API URLs for MCP integrations.
  - MCP metadata now uses the public API address, while server-side tool requests use the internal address.
  - Added the optional `KANEO_INTERNAL_API_URL` variable, defaulting to `http://127.0.0.1:1337`.
  - API paths are handled correctly when the configured URL includes `/api`.

- **Documentation**
  - Documented the new MCP environment variable and server-side HTTP usage.

- **Tests**
  - Added coverage for public metadata URLs and internal authentication requests.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,6 +1,7 @@
 # General URLs
 KANEO_CLIENT_URL=http://localhost:5173
 # KANEO_API_URL=http://localhost:1337      # optional - defaults to KANEO_CLIENT_URL/api
+# KANEO_INTERNAL_API_URL=http://127.0.0.1:1337 # optional - internal MCP tool requests only
 # CORS_ORIGINS=                            # optional - comma-separated; defaults to KANEO_CLIENT_URL
 
 # Database
--- a/ENVIRONMENT_SETUP.md
+++ b/ENVIRONMENT_SETUP.md
@@ -43,6 +43,7 @@ For local development, the web app also supports:
 ### Optional Variables
 
 Kaneo supports many optional configuration options including:
+- `KANEO_INTERNAL_API_URL` - API origin used only for server-side requests from the built-in HTTP MCP endpoint. Defaults to `http://127.0.0.1:1337`; override it only if the API is not reachable there from its own process.
 - SSO providers (GitHub OAuth via `GITHUB_OAUTH_CLIENT_ID` / `GITHUB_OAUTH_CLIENT_SECRET`, Google, Discord, Custom OAuth/OIDC)
 - GitHub repository integration (GitHub App: `GITHUB_APP_ID`, `GITHUB_PRIVATE_KEY`, `GITHUB_WEBHOOK_SECRET`, optional `GITHUB_APP_NAME`), separate from GitHub SSO
 - SMTP configuration for email
--- a/apps/api/src/mcp/index.ts
+++ b/apps/api/src/mcp/index.ts
@@ -28,12 +28,14 @@ import {
 } from "./schemas";
 import { registerMcpTools, toMcpToolRegistrar } from "./tools";
 
-const publicApiUrl = (
-  process.env.KANEO_API_URL || "http://localhost:1337"
-).replace(/\/api\/?$/, "");
+const publicApiUrl = (process.env.KANEO_API_URL || "http://localhost:1337")
+  .replace(/\/api\/?$/, "")
+  .replace(/\/+$/, "");
 const internalApiUrl = (
   process.env.KANEO_INTERNAL_API_URL || "http://127.0.0.1:1337"
-).replace(/\/api\/?$/, "");
+)
+  .replace(/\/api\/?$/, "")
+  .replace(/\/+$/, "");
 
 type McpSession = {
   transport: WebStandardStreamableHTTPServerTransport;
--- a/apps/api/src/mcp/index.ts
+++ b/apps/api/src/mcp/index.ts
@@ -28,10 +28,12 @@ import {
 } from "./schemas";
 import { registerMcpTools, toMcpToolRegistrar } from "./tools";
 
-const apiUrl = (process.env.KANEO_API_URL || "http://localhost:1337").replace(
-  /\/api\/?$/,
-  "",
-);
+const publicApiUrl = (
+  process.env.KANEO_API_URL || "http://localhost:1337"
+).replace(/\/api\/?$/, "");
+const internalApiUrl = (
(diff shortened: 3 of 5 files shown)


tinsever:
Great, pretty much what I had lying around locally. :)

PR #1562, opened 2026-08-11
feat(auth): let admins restrict workspace creation to instance admins

pesnik:
## Description

Right now `allowUserToCreateOrganization: true` in `apps/api/src/auth.ts` lets any signed-in user create a workspace ("organization" internally), with no way to restrict it. There's no existing config flag for this — unlike `DISABLE_REGISTRATION` or `DISABLE_GUEST_ACCESS`, which follow the same pattern for adjacent concerns.

This adds `DISABLE_WORKSPACE_CREATION`: when set, only instance admins (`user.role === "admin"`, the same role the first-registered user auto-gets) may create workspaces. The name and semantics deliberately mirror `DISABLE_REGISTRATION` — the flag names the capability being turned off for ordinary users, not the role that stays exempt, since Kaneo's *per-workspace* roles (`owner`/`admin`/`member`/`viewer`, which can be renamed/customized per workspace via `dynamicAccessControl`) don't exist yet at workspace-creation time. Only the instance-wide role is meaningful at that point.

Also hides the "Add workspace" affordance in the workspace switcher and command palette for non-admins when the flag is set, so they don't hit a raw 403. The server-side check remains the actual enforcement boundary regardless of what the UI shows.

## Verified manually

Against a clean local Postgres + fresh instance:
- First user still bootstraps as instance admin and can create their first workspace via onboarding, unaffected.
- A second, non-admin user invited into that workspace no longer sees "Add workspace" in the workspace switcher dropdown or the `Cmd+K` command palette.
- Attempting the underlying `organization/create` call directly as that non-admin returns `403` ("You are not allowed to create a new organization") — confirmed via the onboarding flow described below.
- With the flag unset (default), behavior is unchanged — any signed-in user can create a workspace, same as today.

## Known limitation (flagging for discussion, not fixed here)

A brand-new **non-admin** user with zero workspaces (e.g. on an instance with open registration but `DISABLE_WORKSPACE_CREATION=true`, and no pending invite) still lands on `/onboarding`'s "Create workspace" form — nothing routes them elsewhere. Submitting it correctly fails server-side with a generic toast ("You are not allowed to create a new organization"), so nothing is broken, but the UX dead-ends them with no path forward. Fixing that well depends on what such a user *should* see (an invite-only message? a "contact your admin" screen?), which felt like a separate product decision out of scope for this change — happy to take a stab at it if maintainers want it in the same PR or a follow-up.

## Type of Change

## How Has This Been Tested?
- [x] Manual testing (see above)
- Typecheck (`pnpm --filter @kaneo/api typecheck`, `pnpm --filter @kaneo/web typecheck`) and existing test suites (`pnpm --filter @kaneo/api test`, `pnpm --filter @kaneo/web test`) pass, diffed against a pre-change baseline so nothing pre-existing is misread as a regression — both suites have 2 pre-existing failures/errors from a missing `@modelcontextprotocol/server` package, byte-identical before and after this change.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added an option to restrict workspace creation to instance administrators.
  * Workspace creation controls, menu items, and keyboard shortcuts now respect this setting.
  * Administrators can continue creating workspaces when restrictions are enabled.


randoneering:
@pesnik thank you for your contribution. Can you please review the suggestions by coderabbit and qodo? Thank you!

pesnik:
@randoneering Pushed a fix addressing the CodeRabbit and Qodo feedback:

- `apps/api/src/auth.ts`: `allowUserToCreateOrganization` now re-reads the user's role from the database instead of trusting the session, which can be cookie-cached and briefly hold the pre-promotion role for the first-user bootstrap — this was causing a spurious 403 when creating the first workspace.
- `command-palette/index.tsx` + `workspace-switcher.tsx`: `canCreateWorkspace` no longer defaults to allowed while the config query is loading/erroring.
- `command-palette/index.tsx`: the "create workspace" keyboard shortcut is now gated the same way the menu item already was, so it can't bypass the hidden affordance.

CodeRabbit has re-reviewed the latest commit (e70d541) and reports no further findings. Qodo hasn't re-run on this commit yet — happy to follow up if it flags anything once it does.

tinsever (approved):
LGTM

PR #1572, opened 2026-08-13
feat: add support for multiple assignees per task (#1334)

### LINKED ISSUE #1334: feat: Add Multiple Assignee in Tasks
### Problem Statement

Currently, Kaneo only supports assigning a single person to a task. The assignee dropdown only allows selecting one user at a time (Unassigned, or one team member), which doesn't reflect how many teams actually work. Tasks often require collaboration between multiple people (e.g., a developer and a reviewer, or a designer and an engineer working together), and right now there's no way to represent that within a single task — users either have to duplicate the task or just pick one person as a workaround, losing visibility into who else is actually involved.

### Proposed Solution

Add support for multiple assignees per task, similar to how Trello handles it. Instead of a single-select dropdown, the assignee field should allow multi-select, showing all assigned avatars on the task card and in the task detail view. Suggested UX is to keep the same dropdown style shown in the current "Unassigned" selector, but switch the selection control to checkboxes so multiple team members can be checked at once and avatars stack/overlap on the task card (like Trello does).

### Alternative Solutions

One alternative is to keep single-assignee but add a "watchers" or "followers" field so additional people get notified without being formally assigned — this is simpler to implement but doesn't give the same clarity of ownership/responsibility that Trello-style multiple assignees provide. Another alternative is to rely on labels or comments to mention other involved people, but this is informal and doesn't show up visually on the task card the way an assignee avatar does.

### Relevant Context

As teams using Kaneo grow, tasks often involve more than one person working together, such as a backend developer pairing with a frontend developer, or a developer working alongside a reviewer. Right now, the single-assignee field forces an artificial choice of who "owns" the task even when the work is genuinely shared between two or more people, which can make it unclear who is actually responsible for what. Supporting multiple assignees would let the task reflect how the work is actually being done, without requiring extra workarounds like duplicate tasks or relying on comments to track who else is involved.

Yes, because the change is additive and opt-in — single-assignee workflows remain exactly as simple as they are today (just select one person), while teams that need to assign multiple people gain that flexibility without added complexity for users who don't need it. The UI change (checkboxes instead of radio-style selection) is minimal and intuitive, since most users are already familiar with multi-select patterns from tools like Trello

ancientdev0x:
## Description
Implements multi-assignee support for tasks end-to-end (#1334).

- **Database**: Introduced `task_assignee` join table schema and migration with automatic backfill from existing `task.assignee_id`.
- **API**: Updated task endpoints (`get-tasks`, `get-task`, `create-task`, `update-task-assignee`, `bulk-update-tasks`) and Hono/Valibot schemas to accept `assigneeIds` and return `assignees` array while maintaining backwards compatibility with `userId`/`assigneeId`.
- **Frontend UI**:
  - Overlapping stacked avatars on task cards on Kanban boards.
  - Multi-select checkbox list in task sidebar assignee popover.
  - Multi-select assignees during initial task creation modal.

Closes #1334

## Summary by CodeRabbit

* **New Features**
  * Tasks can now be assigned to multiple people.
  * Create and update flows support selecting, changing, and removing multiple assignees.
  * Task cards and details display multiple assignee avatars and names.
  * Assignee filters match tasks assigned to any selected person.
  * Notifications are sent to relevant assignees when tasks are created.
  * Existing single-assignee tasks remain supported.


randoneering:
@PriyanshusSGupta again, thank you for your contribution and fix. Please review the coderabbit suggestions when you have a chance!

ancientdev0x:
@randoneering Thank you! All CodeRabbit suggestions have been reviewed, addressed, and verified in the latest commits. All CI checks are passing cleanly.

TymekV:
When right-clicking and adding asignees it shows success but without actually adding them. Also, the selector hides after selecting one asignee.

ancientdev0x:
@TymekV Great catch, thank you! Fixed in the latest commit:
- Updated the right-click context menu to pass the full `assigneeIds` payload when toggling assignees.
- Set `closeOnClick={false}` on assignee submenu items so the selector stays open for selecting/deselecting multiple assignees seamlessly.
- Fixed the toast notification to only fire on actual success rather than in the `finally` block.

Piernikkk:
Why?

TymekV:
Hi, why was this PR closed?

I really hoped that the multiple aasignees feature would be merged in Kaneo.

PR #1575, opened 2026-08-13
fix(time-entry): compute duration and reject an inverted range on create

alooshxl:
## Description

`createTimeEntry` always stored `duration: 0`, even for an entry created complete with an explicit `endTime`.

The controller accepted a `duration` parameter, but nothing ever passed one — the `POST /time-entry` route does not forward it, and the route's Valibot validator does not accept the field at all. `duration` was therefore always `undefined`, so `duration: duration || 0` stored `0` unconditionally. The controller also never checked that `endTime` is after `startTime`, so an inverted range was silently accepted.

Repro: `POST /api/time-entry` with `startTime` `2026-08-10T10:00:00Z` and `endTime` `2026-08-10T11:00:00Z` stores `duration: 0` instead of `3600`. The MCP `create_time_entry` tool goes through the same route, so an entry logged through it reports `0` too, despite the tool documenting "Omit endTime to leave the entry running".

This removes the vestigial parameter, derives the duration from the two timestamps, and adds the inverted-range guard — mirroring `updateTimeEntry`, which was hardened the same way in #1554. The guard reuses that controller's exact 400 message, so both paths respond identically.

### Behaviour change for running entries

An entry created without an `endTime` now stores `duration: null` rather than `0`. That matches `updateTimeEntry`, the nullable column (`integer("duration").default(0)` — no `.notNull()`), and `timeEntrySchema` (`v.nullable(v.number())`). `null` is also the more accurate value: `0` asserts that no time elapsed, `null` says the duration is not known yet.

I checked every consumer of the field before making this change:

- **`apps/web`** — the only reference is the `TimeEntry` type, which is already `duration: number | null`. No component reads the field: the time-entry fetchers and hooks are not imported by any UI yet, and nothing sums or formats the value. (`apps/web/src/lib/format-duration.ts` takes a `number`, but it is not imported anywhere.)
- **`packages/mcp`** — `create_time_entry` sends only `taskId`/`startTime`/`endTime`/`description` and returns the API response verbatim.
- **`apps/api`** — `get-time-entries.ts` selects the column as-is.
- **`tests/api-integration/account-deletion.test.ts`** — inserts a duration straight through the DB, so it does not exercise this path.

No consumer needs a null-safe read, so nothing outside the controller changed.

## Related Issue(s)

No filed issue — found while reading the time-entry controllers alongside #1554.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test addition or update

## How Has This Been Tested?

- [x] Unit tests
- [x] Integration tests
- [ ] Manual testing

New `tests/api/time-entry/create-time-entry.test.ts`, following the mocking conventions in the sibling `update-time-entry.test.ts`:

1. stores `3600` for a 10:00 → 11:00 entry
2. leaves `duration` as `null` for an entry with no `endTime`
3. rejects a `startTime` later than the `endTime` with a 400 and the shared message, without inserting

All three fail against the current code and pass with the fix.

Full local run on the branch:

- `pnpm exec biome ci .` — exit 0 (78 warnings, 2 infos, unchanged from `main`)
- `pnpm typecheck` — 6/6 tasks
- `pnpm test` — API 358 tests / 53 files, web 91 tests / 30 files
- `pnpm build` — 7/7 tasks
- `pnpm test:integration` — 181 tests / 29 files, against Postgres 16

N/A — API-only change.

## Checklist

## Additional Notes

`apps/docs/openapi.json` does not need regenerating. The request shape is unchanged — the removed `duration` parameter was internal and never part of the validator — and the documented response already describes `duration` as `{"type": "number", "nullable": true}`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Time entries now calculate duration automatically from their start and end times.
  * Ongoing entries retain an unset duration until they are completed.
  * Invalid entries with an end time before the start time are rejected.
  * Duration values are preserved correctly when creating time entries.

* **Tests**
  * Added coverage for duration calculation, ongoing entries, and invalid time ranges.

### DIFF
--- a/apps/api/src/time-entry/controllers/create-time-entry.ts
+++ b/apps/api/src/time-entry/controllers/create-time-entry.ts
@@ -11,15 +11,25 @@ async function createTimeEntry({
   description,
   startTime,
   endTime,
-  duration,
 }: {
   taskId: string;
   userId: string;
   description?: string;
   startTime: Date;
   endTime?: Date;
-  duration?: number;
 }) {
+  if (endTime && startTime.getTime() > endTime.getTime()) {
+    throw new HTTPException(400, {
+      message:
+        "Start time cannot be after end time. Please adjust the time range.",
+    });
+  }
+
+  let duration: number | null = null;
+  if (endTime) {
+    duration = Math.floor((endTime.getTime() - startTime.getTime()) / 1000); // duration in seconds
+  }
+
   const [createdTimeEntry] = await db
     .insert(timeEntryTable)
     .values({
@@ -29,7 +39,7 @@ async function createTimeEntry({
       description: description || "",
       startTime,
       endTime: endTime || null,
-      duration: duration || 0,
+      duration,
     })
     .returning();
 
--- a/tests/api/time-entry/create-time-entry.test.ts
+++ b/tests/api/time-entry/create-time-entry.test.ts
@@ -1,0 +1,121 @@
+import { beforeEach, describe, expect, it, type Mock, vi } from "vitest";
+
+const mockInsert = vi.fn();
+const mockSelect = vi.fn();
+const mockPublishEvent = vi.fn();
+
+vi.mock("../../../apps/api/src/database", () => ({
+  default: {
+    insert: (...args: unknown[]) => mockInsert(...args),
+    select: (...args: unknown[]) => mockSelect(...args),
+  },
+}));
+
+vi.mock("../../../apps/api/src/events", () => ({
+  publishEvent: (...args: unknown[]) => mockPublishEvent(...args),
+}));
+
+import createTimeEntry from "../../../apps/api/src/time-entry/controllers/create-time-entry";
+
+function makeInsertMock(createdRow: unknown) {
+  const returning = vi.fn(() => Promise.resolve([createdRow]));
+  const values = vi.fn(() => ({ returning }));
+  return { values, returning };
+}
+
+function makeSelectMock(rows: unknown[]) {
+  const chain: Record<string, Mock> = {
+    from: vi.fn(() => chain),
+    where: vi.fn(() => Promise.resolve(rows)),
+  };
+  return chain;
+}
+
+describe("createTimeEntry", () => {
+  beforeEach(() => {
+    vi.resetAllMocks();
+  });
+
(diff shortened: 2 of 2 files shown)


luantaraschi (approved):
Checked this out at `95b2f372`. The diagnosis and the mirroring both hold.

**The vestigial parameter really was unreachable.** The route's Valibot object at `apps/api/src/time-entry/index.ts:80-85` accepts `taskId`, `startTime`, `endTime` and `description`, and the handler at line 92 forwards exactly those four. So `duration` arrived `undefined` on every call and `duration || 0` wrote `0` even for an entry created complete. Removing the parameter rather than plumbing it through is the right call: nothing could have supplied it without also changing the validator.

**The mirroring is exact.** `updateTimeEntry` already had both halves, and this matches them line for line:

```ts
// update-time-entry.ts:32
"Start time cannot be after end time. Please adjust the time range.",
// update-time-entry.ts:36-41
let duration: number | null = null;
if (effectiveEndTime) {
  duration = Math.floor((effectiveEndTime.getTime() - startTime.getTime()) / 1000);
}
```

Same message, same seconds, same `Math.floor`, same null. Two endpoints that disagree about the same field is exactly the kind of thing that bites later, so I like that this closes the gap rather than inventing a third behaviour.

**On the `0` to `null` change for running entries**, which is the part worth being careful about. It looks safe from where I can see:

- the column is `integer("duration").default(0)` with no `.notNull()`, so the database already allows it
- `apps/web/src/types/time-entry/index.ts:2` already declares `duration: number | null`
- of the seven files in `apps/web` that touch time entries, all are fetchers, hooks and the type; none renders or sums the field yet, so nothing is doing arithmetic that `null` would turn into `NaN`
- `get-time-entries.ts:15` selects the column straight through without a coalesce

The one thing I could not check from here is any consumer outside this repo reading the API and assuming a number. Worth a line in the release notes if you keep them, but I would not hold the PR for it.

**What I ran**, on Windows with Node 22.20.0:

- reverting only `create-time-entry.ts` to `main` and keeping the tests: all 3 fail, so they are measuring the change and not something that was already true
- `vitest run tests/api/time-entry`: 7 passing across 2 files
- `biome check` on both changed files: clean

Looks good to me.

PR #1576, opened 2026-08-13
fix(github): ignore non-branch refs on push webhooks

alooshxl:
## Description

`handlePush` derived the branch name with `ref.replace("refs/heads/", "")`. `String.replace` is a no-op when the needle is absent, so a non-branch ref fell through with `branchName` set to the entire ref.

The handler is registered for the `push` event, and GitHub delivers `push` for tag creation and deletion as well. Nothing upstream (`handleGithubWebhookRoute` → `handleGitHubWebhook` → `verifyAndReceive` → the registered `push` handler) filters the ref, so tag pushes reached the same code path as branch pushes.

Repro: push a tag `v1.0` to a repository with a Kaneo GitHub integration. The webhook arrives with `ref: "refs/tags/v1.0"`; `branchName` becomes the literal string `refs/tags/v1.0`, it is not in `PROTECTED_BRANCHES`, and Kaneo runs the integration lookup and branch matching against that string on every tag push.

Usually this only wastes an integration query, because the default pattern compiles to an anchored regex that `refs/tags/...` cannot match. With a `customBranchRegex`, which is applied unanchored, it is a real state change: a tag `refs/tags/kan-42` matches `kan-(\d+)`, and the task is linked and moved to `in-progress` by a release tag.

This adds the `refs/heads/` guard before stripping and switches `replace` to `slice`, which is exactly what the Gitea push handler already does (`apps/api/src/plugins/gitea/webhooks/push.ts`), so the two providers now read identically. It is the same shape of fix as #1522.

Scope note: GitHub was the only affected handler. Gitea already guards. The pull request handlers read `pull_request.head.ref`, which is a bare branch name with no prefix to strip, and there are no other inbound source-control providers (`generic-webhook` is outbound only).

## Related Issue(s)

No filed issue; found while reading the two push handlers side by side.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test addition or update

## How Has This Been Tested?
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual testing

Adds `tests/api/plugins/github/webhooks/push.test.ts` (8 cases), mocking the task service, link manager, column resolver and event bus, and exercising the real branch matcher:

- returns early for `refs/tags/v1.0`, `refs/tags/v2.3.4`, `refs/pull/12/merge` and `refs/notes/commits`, asserting the integration lookup is never reached
- leaves the task untouched when a tag matches an unanchored `customBranchRegex`
- still looks up integrations for `refs/heads/feature-x`
- still strips the prefix so `refs/heads/kan-42` links `kan-42` and updates the task status
- still skips a protected branch

Reverting the source change fails 5 of the 8 cases; the custom-regex case fails with `updateTaskStatus` called as `("task-1", "in-progress")`, which is the task-mutating behaviour described above. The remaining 3 pass either way and are there to guard the branch path against regression.

Full suite on this branch: `biome ci .` clean (78 warnings, 2 infos, 0 errors, unchanged from `main`), `pnpm typecheck` 6/6, `pnpm test` API 363 tests across 53 files and web 91 across 30, `pnpm build` 7/7.

## Checklist

## Additional Notes

The log prefix uses this file's existing `[Push]` style rather than Gitea's `[Gitea Push]`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GitHub push webhook handling by ignoring tags and other non-branch references.
  * Correctly extracts branch names from valid branch references.
  * Prevents updates for protected branches.

* **Tests**
  * Added comprehensive coverage for branch pushes, tags, pull references, notes, custom tag patterns, task lookups, links, and status updates.

### DIFF
--- a/apps/api/src/plugins/github/webhooks/push.ts
+++ b/apps/api/src/plugins/github/webhooks/push.ts
@@ -36,7 +36,12 @@ const PROTECTED_BRANCHES = [
 export async function handlePush(payload: PushPayload) {
   const { ref, repository, head_commit } = payload;
 
-  const branchName = ref.replace("refs/heads/", "");
+  if (!ref.startsWith("refs/heads/")) {
+    console.log(`[Push] Skipping non-branch ref: ${ref}`);
+    return;
+  }
+
+  const branchName = ref.slice("refs/heads/".length);
   console.log(`[Push] Processing branch: ${branchName}`);
 
   if (PROTECTED_BRANCHES.includes(branchName)) {
--- a/tests/api/plugins/github/webhooks/push.test.ts
+++ b/tests/api/plugins/github/webhooks/push.test.ts
@@ -1,0 +1,168 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const mockFindAllIntegrationsByRepo = vi.fn();
+const mockFindTaskByNumber = vi.fn();
+const mockIsTaskInFinalState = vi.fn();
+const mockUpdateTaskStatus = vi.fn();
+const mockCreateOrUpdateExternalLink = vi.fn();
+const mockResolveTargetStatus = vi.fn();
+const mockPublishEvent = vi.fn();
+
+vi.mock(
+  "../../../../../apps/api/src/plugins/github/services/task-service",
+  () => ({
+    findAllIntegrationsByRepo: (...args: unknown[]) =>
+      mockFindAllIntegrationsByRepo(...args),
+    findTaskByNumber: (...args: unknown[]) => mockFindTaskByNumber(...args),
+    isTaskInFinalState: (...args: unknown[]) => mockIsTaskInFinalState(...args),
+    updateTaskStatus: (...args: unknown[]) => mockUpdateTaskStatus(...args),
+  }),
+);
+
+vi.mock(
+  "../../../../../apps/api/src/plugins/github/services/link-manager",
+  () => ({
+    createOrUpdateExternalLink: (...args: unknown[]) =>
+      mockCreateOrUpdateExternalLink(...args),
+  }),
+);
+
+vi.mock(
+  "../../../../../apps/api/src/plugins/github/utils/resolve-column",
+  () => ({
+    resolveTargetStatus: (...args: unknown[]) =>
+      mockResolveTargetStatus(...args),
+  }),
+);
+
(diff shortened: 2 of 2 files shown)


luantaraschi (approved):
Checked this out at `94fb0a06` and ran it. The diagnosis holds, and the severity note in the description is right for the reason it gives.

**The two claims, measured against the real helpers.** With the old `ref.replace("refs/heads/", "")`, a tag push leaves `branchName` as the whole ref:

```
ref                    : refs/tags/kan-42
branchName before fix  : "refs/tags/kan-42"
default pattern regex  : /^kan-(\d+)(?:-.*)?$/i
default match on tag   : false
custom regex           : /kan-(\d+)/i
custom match on tag    : task #42
```

So the default path really is saved by the `^...$` that `createBranchRegex` adds, and the `customBranchRegex` path really is not: `extractTaskNumberFromBranch` compiles it with `new RegExp(config.customBranchRegex, "i")` and no anchors, so a release tag lands on the task and moves it to `in-progress`. That is a state change from a tag push, which is worth fixing rather than just the wasted query.

**The tests do their job.** Reverting only `push.ts` to `main` and keeping the test file, 5 of the 8 fail, including `leaves a task untouched when a tag matches an unanchored custom regex`. The other 3 pass on both sides, which is what you want from the branch-ref controls.

**The strongest argument for the shape is already in the repo.** `apps/api/src/plugins/gitea/webhooks/push.ts` on `main` has the same guard, down to the `.slice("refs/heads/".length)` and the log line:

```ts
if (!ref.startsWith("refs/heads/")) {
  console.log(`[Gitea Push] Skipping non-branch ref: ${ref}`);
  return;
}
```

So this is not a new convention, it brings the GitHub handler in line with its Gitea twin that was already correct. The guard also sits before the integration lookup, which is where the Gitea one sits too.

**What else I ran**, on Windows with Node 22.20.0:

- `vitest run tests/api/plugins/github`: 49 passing across 7 files
- `biome check` on both changed files: clean
- `tsc --noEmit` in `apps/api`: 8 errors, all in files this diff does not touch (`src/auth.ts` and the unbuilt `@kaneo/email` / `@kaneo/permissions` / `@modelcontextprotocol/server` imports), so none of them belongs to this change

Nothing blocking from me. Nice catch, and thank you for writing the repro in terms of what a user would actually do.

Issue #1580, opened 2026-08-13
[Bug]: Subtask to parent navigation throws error


janbiasi:
### Bug Description

If you're on a subtask and try to navigate to the parent via the `Subtask of ...` link, the page will simply throw an error in full-screen:

```
Something went wrong!
Cannot read properties of null (reading 'commands')
```

This is reproducible and at least occurs since 2.16.x (not sure if it already occured earlier.
Current workaround is to just hard-reload the task page.

### Steps To Reproduce

1. Navigate to a task with subtasks
2. Click on the subtask
3. Click on the top link `Subtask of ...`
4. Error occurs on the task detail page

### Expected Behavior

1. Navigate to a task with subtasks
2. Click on the subtask
3. Click on the top link `Subtask of ...`
4. Back on the parent task (1.)

### Actual result

```
Something went wrong!
Cannot read properties of null (reading 'commands')
```

### Environment

- Kaneo Version: 2.16.4 (maybe earlier) to latest (currently running 2.18.0)
- Deployment method: Docker 
- Browser: Tested on Helium 0.15.3.1, Chrome 150.0.7871.129, Samsung Browser on OneUI 8.5
- OS: macOS Tahoe and Android 16

hydraxman:
Confirmed on current `main` (`3fc77f00`): rerendering `TaskDescription` from a child task ID to its parent reproduces the reported `Cannot read properties of null (reading 'commands')` failure.

The task ID is captured by `handleAssetFileUpload`, and that callback is a `useEditor` dependency. Navigating to another task therefore destroys/recreates the Tiptap editor while the description-sync effect still receives the destroyed instance and reads `editor.commands`.

Implementation plan:
- keep the upload callback stable across task navigation by reading the current task ID from a ref;
- preserve the existing upload destination (the currently displayed task);
- add a component regression that renders a child description, rerenders with the parent task ID, and verifies the parent description without an editor lifecycle error.

Alternatives considered: an `isDestroyed` guard would avoid the exception but still discard/recreate the editor unnecessarily; keying the editor per task would make that reset explicit rather than fixing the unstable dependency.

Scope is limited to task-description editor lifecycle and its component test; no API, schema, or relation behavior changes. Acceptance is that child-to-parent navigation renders the parent description without a full-screen error and uploads after navigation still target the active task.

ab192130:
This bug is fixed by **https://github.com/usekaneo/kaneo/pull/1581**.

[...]

PR #1587, opened 2026-08-15
fix(web): make pending task description updates reliable across task switches

ab192130:
## What

Hardening for the task description save pipeline, split out of #1581 per review feedback. This PR makes pending description updates reliable when the user switches tasks or closes the page mid-edit.

## Changes

1. **Bind pending description updates to their source task** — the 700 ms debounced update captures the task id + snapshot at edit time, so navigating away mid-debounce can never send one task's markdown to another task's endpoint.
2. **Keep per-task pending updates independent** — each task gets its own debounce timer, so editing task B within the interval no longer cancels task A's pending save.
3. **Persist fresh task state at save time** — when the timer fires, the update uses the freshest task snapshot for that task, so field changes made during the 700 ms window (status, assignee, due date) aren't overwritten by the older copy.
4. **Flush pending updates on unmount** — instead of just cancelling timers, the cleanup now runs any pending save immediately, so an edit typed right before closing the view isn't silently lost (addresses review feedback to replace clear with flush).
5. **Resolve the task at save time** — the debounce no longer requires the task to be loaded at keystroke time; the update resolves the task (id-matched fresh snapshot, else the captured one) when the timer fires, so a character typed during the initial task fetch is no longer dropped.

## Verification

- `pnpm --filter @kaneo/web typecheck` — pass
- `pnpm --filter @kaneo/web test` — task description test passes (pre-existing suite failures in `env.test.ts`/`use-board-sort`/`use-task-filters` are jsdom-env issues unrelated to this change)

Depends on #1581 (the core editor-lifecycle fix).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task description updates to reliably save the latest changes.
  * Prevented updates from being applied to the wrong task.
  * Ensured pending description edits are saved when leaving the editor.
  * Added clearer error logging when a description update fails.

### DIFF
--- a/apps/web/src/components/task/task-description.tsx
+++ b/apps/web/src/components/task/task-description.tsx
@@ -77,6 +77,7 @@ import { isInCodeBlockLanguagePicker } from "@/lib/is-in-codeblock-language-pick
 import { getSharedShikiHighlighter } from "@/lib/shiki-highlighter";
 import { toast } from "@/lib/toast";
 import { uploadTaskImage } from "@/lib/upload-task-image";
+import type Task from "@/types/task";
 import { AttachmentCard } from "./extensions/attachment-card";
 import { EmbedBlock } from "./extensions/embed-block";
 import { KaneoIssueLink } from "./extensions/kaneo-issue-link";
@@ -576,22 +577,26 @@ export default function TaskDescription({ taskId }: TaskDescriptionProps) {
   }, []);
 
   const debouncedUpdate = useCallback(
-    debounce(async (markdown: string) => {
-      if (!canEditRef.current) return;
+    debounce(
+      async (
+        markdown: string,
+        taskId: string | undefined,
+        currentTask: Task | undefined,
+      ) => {
+        const updateTaskFn = updateTaskRef.current;
+        if (!taskId || !currentTask || !updateTaskFn) return;
 
-      const currentTask = taskRef.current;
-      const updateTaskFn = updateTaskRef.current;
-      if (!currentTask || !updateTaskFn) return;
-
-      try {
-        await updateTaskFn({
-          ...currentTask,
-          description: markdown,
-        });
-      } catch (error) {
-        console.error("Failed to update description:", error);
-      }
-    }, 700),
+        try {
+          await updateTaskFn({
+            ...currentTask,
+            description: markdown,
+          });
+        } catch (error) {
+          console.error("Failed to update description:", error);
+        }
+      },
+      700,
+    ),
     [],
   );
 
@@ -806,7 +811,7 @@ export default function TaskDescription({ taskId }: TaskDescriptionProps) {
         const markdown = formatMarkdown(activeEditor.getMarkdown());
         if (markdown === latestSyncedMarkdownRef.current) return;
         latestSyncedMarkdownRef.current = markdown;
-        debouncedUpdate(markdown);
+        debouncedUpdate(markdown, taskIdRef.current, taskRef.current);
       },
     },
     [getOverlayPosition, handleAssetFileUpload, t, toShikiLanguage],
--- a/apps/web/src/components/task/task-description.tsx
+++ b/apps/web/src/components/task/task-description.tsx
@@ -595,9 +595,11 @@ export default function TaskDescription({ taskId }: TaskDescriptionProps) {
         void (async () => {
           const updateTaskFn = updateTaskRef.current;
           if (!updateTaskFn) return;
+          const latest = taskRef.current;
+          const base = latest?.id === taskId ? latest : currentTask;
           try {
             await updateTaskFn({
-              ...currentTask,
+              ...base,
               description: markdown,
             });
           } catch (error) {
--- a/apps/web/src/components/task/task-description.tsx
+++ b/apps/web/src/components/task/task-description.tsx
@@ -65,7 +65,6 @@ import { useUpdateTaskDescription } from "@/hooks/mutations/task/use-update-task
 import useGetTask from "@/hooks/queries/task/use-get-task";
 import { useWorkspacePermission } from "@/hooks/use-workspace-permission";
 import { cn } from "@/lib/cn";
-import debounce from "@/lib/debounce";
 import { parseTaskListMarkdownToNodes } from "@/lib/editor-task-list-paste";
 import {
   extractIssueKeyFromUrl,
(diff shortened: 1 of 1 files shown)


luantaraschi (commented):
All three points from the earlier review land here, and I could measure each one instead of reading it. I drove the component through the `useEditor` mock seam that #1581 introduced, with fake timers, comparing this branch against `main`.

```
main:   saved id = task-b, description = "typed into task A"  |  unmount: mutateAsync calls = 0
#1587:  saved id = task-a, description = "typed into task A"  |  unmount: mutateAsync calls = 1
```

Typing in A, switching to B on the full page route and letting the timer fire now writes under A. Typing and closing the sheet within 700 ms now saves instead of dropping the edit. A keystroke typed before `useGetTask` resolves still lands, which was the `!currentTask` case.

`saveDescriptionUpdate` resolving with `latest?.id === taskId ? latest : task` is the right shape: it takes the fresh row when the ref still points at the same task, so a status or assignee change inside the window is not overwritten, and falls back to the captured snapshot only when it cannot. Keying the timers by task id rather than using one shared timer means two tasks edited within the same window both save.

## The thing I would push back on

**This PR has no tests.** It rewrites the whole save path, and the diff against #1581 is 127 lines in `task-description.tsx` and zero in `task-description.test.tsx`. Every behaviour above is a regression waiting to happen, and each one is three lines to pin with the seam already in the file.

The three I ran, if they are useful as a starting point:

```tsx
it("saves under the task the keystroke belongs to", () => {
  mocks.tasks.set("task-a", makeTask("task-a", "body A"));
  mocks.tasks.set("task-b", makeTask("task-b", "body B"));

  const { rerender } = render(<TaskDescription taskId="task-a" />);
  capturedOnUpdate?.({ editor: editorStub("typed into task A") });

  rerender(<TaskDescription taskId="task-b" />);
  vi.advanceTimersByTime(700);

  const saved = mocks.mutateAsync.mock.calls[0]?.[0];
  expect(saved?.id).toBe("task-a");
  expect(saved?.description).toBe("typed into task A");
});

it("flushes the pending save on unmount instead of dropping it", () => {
  mocks.tasks.set("task-a", makeTask("task-a", "body A"));
  const { unmount } = render(<TaskDescription taskId="task-a" />);
  capturedOnUpdate?.({ editor: editorStub("typed then closed") });
  unmount();
  expect(mocks.mutateAsync).toHaveBeenCalledTimes(1);
});

it("keeps a keystroke typed before the task resolves", () => {
  const { rerender } = render(<TaskDescription taskId="task-a" />);
  capturedOnUpdate?.({ editor: editorStub("typed during the fetch") });
  mocks.tasks.set("task-a", makeTask("task-a", "body A"));
  rerender(<TaskDescription taskId="task-a" />);
  vi.advanceTimersByTime(700);
  expect(mocks.mutateAsync).toHaveBeenCalledTimes(1);
});
```

They need `useGetTask` mocked to read from a map keyed by task id, `useUpdateTaskDescription` returning a `mutateAsync` spy, and `useEditor` capturing `options.onUpdate`. The rest of the mock block is what `task-description.test.tsx` already has. All three fail on `main` for the first two and pass here, so they measure the change rather than restating it.

## Smaller notes

`@/lib/debounce` is still imported by `task-title.tsx` and `mermaid-block.ts`, so dropping it here does not orphan the module.

This branch and #1588 both build on #1581 and both touch `task-description.tsx`, in the `handleAssetFileUpload` region and the file input handler. Whichever lands second will need a rebase. Worth deciding the order now rather than at merge time.

Web suite here: 84 tests pass across 29 files, same as `main`. The two files that fail to load are `@kaneo/permissions` not being built in my install, and they fail the same way on `main`.

tinsever:
This PR should not be merged or reviewed further in its current form.

It addresses a separate, pre-existing autosave problem that is not part of #1580, but it has no dedicated issue or acceptance criteria and rewrites the description-save pipeline without adding tests for any of the new behavior.
The current implementation also does not satisfy its own reliability claims. If an edit is made before the task has loaded and the 700 ms timer expires first, the pending entry is removed before saveDescriptionUpdate discovers that no task snapshot is available. The function then returns and the edit is permanently lost.

There are additional untested state transitions, including returning to a task while its previous edit is still pending. Hydrating the cached server description and editing again can replace that pending markdown. The unmount cleanup may start a mutation during a React unmount, but it does not guarantee that an edit is persisted when the browser page is closed.

Please move this back to draft and:
- rebase after #1581 is finalized;
- simplify the state model;
- add behavioral tests for task attribution, slow initial loading, rapid A > B > A navigation, multiple pending tasks, unmounting, and failed mutations.

Passing typecheck and manually probing selected paths is not sufficient for an asynchronous save pipeline that can lose or misattribute user data. The author needs to establish these guarantees before requesting another core-team review.

luantaraschi:
@tinsever is right about the slow initial load, and my earlier comment was wrong on that point. I said a keystroke typed before `useGetTask` resolves still lands. What I actually measured was the ordering where the task arrives inside the 700 ms window. The other ordering loses the edit, and I have now measured both:

```
timer fired while loading: calls right after the timer = 0, calls after the task arrives = 0
task arrived within 700ms:  calls = 1
```

The first case is permanent. `timers.delete(taskId)` runs before `saveDescriptionUpdate`, so by the time it finds neither `taskRef.current` nor the captured snapshot and returns, the entry is already gone and nothing retries when the task does arrive.

One thing worth separating, since it changes what this PR owes. The same probe on `main`:

```
main: timer fired while loading: calls right after the timer = 0, calls after the task arrives = 0
```

Identical, because `main` returns on `!currentTask` at flush time and drops the timer just the same. So this is not a regression introduced here, it is a case the PR describes as fixed and does not fix. Reordering so the entry is only deleted once a save is actually dispatched, or keeping it pending until a task resolves, would close it.

Sorry for the sloppy generalisation from one ordering. The three behaviours I did measure still hold, and the probes for them are in my earlier comment, but this one should go in the test list too.

ab192130:
﻿Rebased onto `main` now that #1581 is merged. The only conflict was the `debouncedUpdate` signature introduced by the bind-to-source change, resolved to the per-task timer design. The branch now carries main's `addToHistory` hydration fix on top of these four commits:

- bind pending description updates to their source task
- keep per-task pending description updates independent
- persist fresh task state for pending description updates
- flush pending description updates on unmount and resolve task at save time

Web typecheck and `task-description.test.tsx` (the real-editor tests from #1595) pass.

PR #1588, opened 2026-08-15
fix(web): surface upload feedback in the task description editor

ab192130:
## What

Upload-feedback UX for the task description editor, split out of #1581 per review feedback.

## Changes

1. **Show a toast when an upload is discarded after a task change** — if the user starts an upload on task A and the task changes (or the editor is torn down) before the upload finishes, the loading toast is dismissed and replaced with `tasks:detail.editor.upload.taskChanged` ("Task changed before the upload finished; file not added"), so the dropped upload isn't silent. The key is added in `en-US.json` and translated in every locale file.
2. **Fall back to the last known editor for early file picks** — clicking the paperclip before the editor has been produced by `immediatelyRender: false` no longer shows the misleading "task changed" toast; it falls back to the last known editor (the pre-existing behavior).

## Verification

- `pnpm --filter @kaneo/web typecheck` — pass
- `pnpm --filter @kaneo/web test` — task description test passes
- `pnpm i18n:check` — all locales OK

Depends on #1581 (the core editor-lifecycle fix).

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented attachments from being added to the wrong task when switching tasks during an upload.
  - Added a clear error message when an upload is discarded because the task changed.
  - Improved task description saving and editor state handling during task changes.

- **Localization**
  - Added upload conflict messages across supported languages.
  - Updated translation validation for task editor and related settings messages.

- **Tests**
  - Added regression coverage for task navigation during image uploads.


luantaraschi (commented):
The i18n side holds up, which is the part I expected to need the most checking.

All seventeen locales carry `tasks:detail.editor.upload.taskChanged` with a real translation rather than an English placeholder:

```
en-US: Task changed before the upload finished; file not added
es-ES: La tarea cambió antes de que terminara la carga; el archivo no se agregó
ru-RU: Задача изменилась до завершения загрузки; файл не добавлен
zh-CN: 上传完成前任务已更改；文件未添加
hi-IN: अपलोड समाप्त होने से पहले कार्य बदल गया; फ़ाइल नहीं जोड़ी गई
```

`i18n/schema.json` matches what `scripts/i18n/schema.mjs` produces, compared as parsed JSON rather than by diff, so it was generated and not hand-edited. And `node ./scripts/i18n/check.mjs` gives byte-identical output on this branch and on `main`, so the new key adds no warning. The plural warnings it does print are already there on `main`.

## One path still drops a file without saying so

The `!pendingInsert` branch restores the `lastEditorRef.current` fallback, but it is guarded:

```js
if (!pendingInsert) {
  if (lastEditorRef.current) {
    void handleAssetFileUpload(file, lastEditorRef.current);
  }
  return;
}
```

`pendingImageInsertRef` is only null when `openImagePicker` was called with no editor, which is the `immediatelyRender: false` window before the editor exists. In that same window `lastEditorRef.current` is usually null too, because it is only assigned in the hydration effect once there is an editor. So the common shape of this case is: picker opens early, user picks a file, both refs are empty, the file is dropped silently. That is the behaviour this PR exists to remove, just on a rarer path.

The fallback also cannot check the task, since the branch is reached precisely because no task id was recorded. Open the picker on A, switch to B, pick the file, and it uploads to B.

Recording the id even when there is no editor would close both:

```js
pendingImageInsertRef.current = {
  editor: activeEditor ?? null,
  range,
  taskId: taskIdRef.current,
};
```

Then the null-editor case can fall back to `lastEditorRef.current`, compare the task id like every other path, and show the same toast when either check fails.

## No tests here either

The two behaviours added are "a discarded upload tells the user why" and "an early file pick still reaches an editor". Both are reachable through the mock seam #1581 added: mock `@/lib/toast` and assert `toast.error` was called with the key, and mock `uploadTaskImage` to assert the fallback path calls it. That is the whole check.

Web suite on this branch: 84 tests pass across 29 files, same as `main` and the same two `@kaneo/permissions` load failures from my install.

One coordination note: this and #1587 both build on #1581 and both edit `task-description.tsx` in the `handleAssetFileUpload` region and the file input handler. Whichever lands second needs a rebase.

tinsever:
Thanks for following up, but I am not going to pursue this as a separate pull request.

The upload toast was not part of #1580 or an accepted product requirement. An optional review suggestion has expanded into another dependent PR with 20 changed files, generated translations, unrelated schema churn, and no behavioral tests for the functionality it adds.

The fallback also loses the originating task identity when pendingInsert is missing. Opening the picker on task A, navigating to task B, and then choosing the file can upload it using the editor for B—the same ownership boundary this work is intended to protect.

There is no need for additional follow-up PRs from review suggestions unless maintainers explicitly asks for one. If we decide that discarded uploads need dedicated feedback, we will track that as its own issue with agreed behavior.

luantaraschi:
Fair, and the framing was mine. The toast started as an aside in my review of #1581, filed under "one smaller note" about a path that drops a file with no feedback. It was an observation, not a request, and I should have said so instead of leaving it to be read as work.

Agreed on the fallback losing the task identity, that is the part I would not want landing either.

Noted on the follow-up PRs.

ab192130:
﻿Rebased onto `main` (on top of the rebased #1587). Two changes:

- The original two commits (task-changed toast + i18n, paperclip fallback to the last editor).
- A new commit: `test(web): cover stale upload completion after task switch` — the coverage @andrejsshell flagged in #1595 as belonging to #1588. It starts an upload on task A, navigates to task B while it is pending, resolves the upload, and asserts the asset is not inserted into B's editor and the `upload.taskChanged` toast fires. Verified to fail if the task-identity guard is removed.

Web typecheck and `task-description.test.tsx` pass.

PR #1610, opened 2026-08-18
feat(sentry): provision alerts and dashboards as code

randoneering:
## Description

Sentry alerts and dashboards are now in the repo as code. The Sentry MCP only reads — writing requires the REST API, so keeping the spec in the repo makes the rules auditable and reproducible.

Six alerts and four dashboards cover the basics: first-seen issues, error spikes, missed crons, slow latency, MCP errors, plus views for the API, web, MCP, and cron monitors.

Two bash scripts provision them. Both are idempotent — they look up by name and update in place, so editing the spec just propagates.

## Related Issue(s)

Fixes #

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test addition or update
- [x] Other (please describe): observability tooling (Sentry alerts and dashboards)

## How Has This Been Tested?

- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing
- [x] Other (please describe): provision scripts ran end-to-end against live Sentry. 5 alerts and 4 dashboards landed. The cron monitor alert waits for cron jobs to run post-deploy — documented in the additional notes below.

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works

## Additional Notes

The cron "missed check-in" alert depends on the four `Sentry.captureCheckIn` monitors existing in Sentry. They appear after the first cron run post-deploy. Re-run the alerts script after the next cron tick to pick them up.

To apply after merge:

```bash
SENTRY_API_TOKEN=sntrys_... ./scripts/provision-sentry-alerts.sh
SENTRY_API_TOKEN=sntrys_... ./scripts/provision-sentry-dashboards.sh
```

## Summary by CodeRabbit

* **New Features**
  * Added automated Sentry alert provisioning for issue, metric, and missed-check-in alerts.
  * Added automated Sentry dashboard provisioning for backend, frontend, MCP, and cron-monitor insights.
  * Added dry-run support, validation, error reporting, and create/update handling for monitoring resources.
  * Added configuration for API errors, latency, web vitals, tool usage, request volume, and monitor health.

* **Documentation**
  * Documented the optional Sentry API token configuration and required access scope.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -82,7 +82,7 @@ SMTP_REQUIRE_TLS=
 # SENTRY_TRACES_SAMPLE_RATE=0.1         # 0.0 disables; default 0 (opt-in)
 # SENTRY_PROFILES_SAMPLE_RATE=0.1      # 0.0 disables; default 0 (opt-in)
 # SENTRY_RELEASE=                       # git SHA; ties events to source maps. Falls back to package.json version.
-# SENTRY_API_TOKEN=                     # Used by scripts/provision-sentry-alerts.sh. Scope: org:write. https://sentry.io/settings/account/api/auth-tokens/
+# SENTRY_API_TOKEN=                     # Used by scripts/provision-sentry-alerts.sh and scripts/provision-sentry-dashboards.sh. Scope: org:write. https://sentry.io/settings/account/api/auth-tokens/
 
 # Kaneo Cloud billing via Creem (Optional): only active when KANEO_CLOUD=true
 # and both keys are set. Self-hosted installs leave these empty; billing stays off.
--- a/.env.sample
+++ b/.env.sample
@@ -82,6 +82,7 @@ SMTP_REQUIRE_TLS=
 # SENTRY_TRACES_SAMPLE_RATE=0.1         # 0.0 disables; default 0 (opt-in)
 # SENTRY_PROFILES_SAMPLE_RATE=0.1      # 0.0 disables; default 0 (opt-in)
 # SENTRY_RELEASE=                       # git SHA; ties events to source maps. Falls back to package.json version.
+# SENTRY_API_TOKEN=                     # Used by scripts/provision-sentry-alerts.sh. Scope: org:write. https://sentry.io/settings/account/api/auth-tokens/
 
 # Kaneo Cloud billing via Creem (Optional): only active when KANEO_CLOUD=true
 # and both keys are set. Self-hosted installs leave these empty; billing stays off.
--- a/scripts/provision-sentry-alerts.sh
+++ b/scripts/provision-sentry-alerts.sh
@@ -21,14 +21,28 @@
 #   - v2: alerts were create-only; edited spec never propagated. now uses PUT.
 set -euo pipefail
 
-ORG="kaneo"
-REGION="de"
-API_BASE="${SENTRY_API_BASE:-https://${REGION}.sentry.io/api/0}"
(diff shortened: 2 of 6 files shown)


randoneering:
Pushed fix `3f55456b`. The first run reported `[ok] Created workflow` for every rule but every API call had silently returned HTTP 400. Three bugs in v1:

- `local + $(...)` was swallowing curl failures (subshell + `set -e` don't mix)
- hidden error response body — couldn't tell where the API rejected
- false success on empty response (jq gets null, .id is empty)

Plus two payload-shape issues that the error body will likely call out:

- workflows were missing `config.frequency` (default now 30 minutes)
- metric-alert workflows carried an empty triggers alongside detectorIds; a workflow is one or the other

The retry path is idempotent: any alerts that did get created on the first run are skipped by name lookup, so re-running just fixes the failed ones. Please re-run:

```bash
SENTRY_API_TOKEN=sntrys_... ./scripts/provision-sentry-alerts.sh
```

The re-run will print the actual HTTP 400 response body for any rule that still fails, which is the signal we need to fix the remaining payload shapes.

randoneering:
Pushed fix `e43550ff`. The actual 400 response bodies on the third run made the payload shape clear:

- issue alerts: missing `conditionResult` on each condition (now `true`)
- metric detectors: missing `eventTypes` on each dataSource and missing top-level `conditionGroup`
- cron alert: the trigger type `monitor_check_in_failure` doesn't exist in the API. Cron alerts fire via detectorIds alone, so the workflow is now just name + enabled (the detector's failure is the trigger).
- GET queries with spaces (`?query=Kaneo API: New issue`) returned HTTP 000. The script now URL-encodes via jq's `@uri` filter.

Re-run with the same token. The retry path is idempotent, so any partial state from the previous runs is skipped by name. The only outstanding dependency is the cron monitors needing at least one tick to exist in Sentry.

randoneering:
Pushed `884de66c`. Two real fixes from the api response bodies:

**1. Sentry migrated transactions to span dataset.** The slow p95 latency and mcp error rate detectors now use `dataset: events_analytics_platform` with `is_transaction:true` in the query. The error spike alert (events dataset, error events) is unaffected.

**2. Metric workflows need a `triggers` field.** The error spike workflow POST returned HTTP 500 because the minimal payload had no `triggers` field. Every Sentry workflow response includes one — even detector-only ones. Now set to `{ logicType: 'any-short', conditions: [] }` so the resource shape matches the api's expectation.

The two issue alerts (workflow 766686, 766687) were updated successfully on this run. The error spike detector (1658482) was created but the workflow POST failed, so it's orphaned. The next run will create a new detector and reference it from the workflow — the orphan can be deleted manually in the Sentry UI.

randoneering:
Two of the update calls didn't actually fix the project scope. Kaneo API: New issue (766686) and Kaneo Web: New issue (766687) were created in an earlier run without a projects field. The latest run sends projects ["kaneo-api"] and projects: ["kaneo-web"] in the payload, and the script reports "Updated workflow" because the PUT returns 200, but Sentry's API silently rejects the empty-to-scoped transition on the projects field. Both alerts still match everything in the org. Confirmed in the UI: both show "No projects connected."

 Fix: delete both alerts in the Sentry UI, then re-run the script. The script will POST them fresh with the projects field set. After that, normal runs work correctly.

 The code is correct for a fresh org. No change needed.

PR #1631, opened 2026-08-20
fix(api): gzip API responses to shrink large board payloads

### LINKED ISSUE #1630: Board fetch (GET /task/tasks/:projectId) ships full labels/externalLinks for every task, unpaginated — multi-MB payloads on large boards
## Summary

The board view fetch (`GET /task/tasks/:projectId`) returns the full, unpaginated task list with every task's labels and external links inlined. On boards with a few thousand tasks, this makes the response multiple megabytes and takes 1-2+ seconds to serialize/transfer, even though the DB query itself is fast.

## Repro / measured impact

Project with 2,802 tasks / 19,512 activity log entries / 3,708 labels:

- `GET /project/:id` → 1.29 MB response, 1.2s
- `GET /task/tasks/:projectId` → 1.31 MB response, 1.7s

Postgres itself responds in single-digit ms — the time is almost entirely payload size (JSON serialization + network transfer), not query latency.

## Root cause

`apps/api/src/task/controllers/get-tasks.ts` already supports pagination (added in #1065):

```ts
const usePagination = options.page != null || options.limit != null;
```

But `apps/web/src/fetchers/task/get-tasks.ts` never passes `page`/`limit`:

```ts
async function getTasks(projectId: string) {
  const response = await client.task.tasks[":projectId"].$get({
    param: { projectId },
  });
  ...
}
```

So the board view always takes the `usePagination = false` path and fetches every task, with every task's full `labels` and `externalLinks` arrays inlined, regardless of board size.

## Suggested fix

Naive page/limit pagination doesn't map cleanly onto a kanban board (you need all cards across all columns visible at once), so the real fix is probably payload shape, not page slicing:

- Split the board-list response into a lightweight summary per task (id, title, status, priority, assignee, label *count* or label chips only) needed to render a card.
- Defer full label objects / external links / description to a per-task detail fetch, triggered when a card is opened.

Happy to put up a PR for this if it's a direction the maintainers are open to — wanted to check the shape of the fix first since it touches the board's data contract on both API and web.

Kampe:
## Summary

Fixes #1630. Enables gzip compression (`hono/compress`) on API responses. Board/task-list endpoints inline full `labels` and `externalLinks` per task with no pagination on the primary board fetch, so boards with a few thousand tasks return multi-MB JSON — every field is genuinely used by the board card UI (label chips, PR status badges), so trimming the response shape isn't a safe option without a larger UX change. Compression is a small, safe win at the transport layer instead.

## Measured impact

Seeded a local board matching a real-world large-board case (2,802 tasks, 2,805 labels) and compared `GET /task/tasks/:projectId`:

| | Before | After |
|---|---|---|
| Response size | 1,730,391 bytes | 59,121 bytes |
| Reduction | — | **96.6%** |

Verified the gzip and non-gzip (`Accept-Encoding: identity`) responses are byte-for-byte identical after decompression — no data/shape change, transport only.

## Test plan

- [x] `pnpm test:unit` — 380/381 passing (pre-existing unrelated failure: an S3 credential test picks up ambient `AWS_PROFILE` from the local shell; fails identically on `main` without this change)
- [x] Manual repro: seeded Postgres directly with 2,802 tasks / 2,805 labels, ran the API locally, compared `curl` with/without `Accept-Encoding: gzip`
- [x] Diffed decompressed gzip body against the uncompressed body — identical

## Summary by CodeRabbit

* **Performance**
  * API responses are now compressed when supported, improving transfer speeds and reducing bandwidth usage.

### DIFF
--- a/apps/api/src/index.ts
+++ b/apps/api/src/index.ts
@@ -9,6 +9,7 @@ import type { Session, User } from "better-auth/types";
 import { eq, sql } from "drizzle-orm";
 import { migrate } from "drizzle-orm/node-postgres/migrator";
 import { Hono } from "hono";
+import { compress } from "hono/compress";
 import { cors } from "hono/cors";
 import { HTTPException } from "hono/http-exception";
 import {
@@ -203,6 +204,11 @@ export function createApp() {
       },
     }),
   );
+
+  // Large boards return multi-MB JSON (board/task list responses embed
+  // labels and external links per task); gzip cuts that by 85-95% since
+  // JSON with repeated keys compresses extremely well.
+  app.use(compress());
 
   const api = new Hono<ApiVariables>();
 


Kampe:
Checked this — `biome.json` sets `javascript.formatter.indentStyle: "space"`, which overrides the top-level `formatter.indentStyle: "tab"` default for TS/JS files specifically (JSON/YAML etc. stay on the top-level tab default). Ran `biome format --write` and `biome check --write` against `apps/api/src/index.ts` locally — zero fixes applied, so the added lines already match the project's actual enforced style. The review bot's guideline appears to be a generic default that doesn't account for this repo's override.

PR #1646, opened 2026-08-22
fix(api): close five workspace-scoping gaps

andrejsshell:
## Description

Closes five places where a workspace boundary was not enforced. Each was found by reading the routes against their own siblings, and each is independent of the others.

**`GET /github-integration/repositories` and `POST /verify` had no workspace scoping** — only the global authenticate middleware. `listUserRepositories()` paginates every repository of every installation of the shared GitHub App, so any authenticated account (including a guest when `DISABLE_GUEST_ACCESS` is unset) could enumerate repository names, owners and private flags across every tenant on the instance, and probe arbitrary owner/repo pairs through `verify`. Both now take a `projectId` and require `workspace:manage_settings`, matching the Gitea equivalents and every other route in the same file. `repositories` becomes `GET /repositories/:projectId`; the `workspaceAccess` lookup deliberately refuses ids from the query string, so a path param is required, validated with Valibot like its siblings. `app-info` is left alone — it returns only the configured app name.

> **This narrows the exposure but does not eliminate it.** The `projectId` gates *who may call* the route; it does not scope *what comes back*. `listUserRepositories()` still iterates `apps.listInstallations()` and returns repositories from every installation of the shared App, so a user holding `workspace:manage_settings` in **any** workspace can still see repository names, owners and private flags belonging to other tenants on the same instance. Closing that properly means listing installations for the calling user's GitHub identity rather than for the App, which is a change to how the integration identifies tenants and does not belong in this PR. Flagged by both reviewers and worth its own issue.

**`getPublicProject` authorized after loading.** It called `getTasks(id)` first — the entire board, every task, label and external link — and only then checked `isPublic`. On an unauthenticated route. The flag is now read on its own before the board is loaded.

**Task assignees were never checked against the workspace.** `createTask` only confirmed the user row existed; `updateTaskAssignee` checked nothing. Any member could assign work to arbitrary accounts on the instance, who then received notifications for a workspace they cannot open. Both go through `assertAssignableUser`, which mirrors `validateWorkspaceAccess` in still allowing instance admins. This also removes the `where id = ''` query `createTask` issued on every unassigned creation.

**`POST /activity/create` took `userId` from the request body**, so any member with `task:update` could write history attributed to someone else. The actor is now the session.

[...]

PR #1647, opened 2026-08-22
fix(api): correct five defects that fail silently

andrejsshell:
## Description

Five defects that produce wrong data or a wrong signal without ever raising an error. None of them fails loudly, which is why the existing suite is green over all of them.

**Time entries persisted zero duration.** `createTimeEntry` wrote `duration: duration || 0` and the `POST /time-entry` route never supplied a duration — only `updateTimeEntry` ever computed one. An entry created with both a start and an end time therefore reported no tracked time until somebody edited it. Duration is now derived from the timestamps on the same formula `updateTimeEntry` uses, and stays `null` while an entry is open rather than claiming zero. It also rejects an end time before the start time, which `updateTimeEntry` already did — otherwise deriving the value would persist a negative duration.

Existing rows are repaired by `backfillTimeEntryDurations()`, wired into `runStartupTasks()` next to the other idempotent backfills. It recomputes closed entries and normalises open ones, touching only rows whose stored duration is `0` or `NULL` *and* whose recomputed value differs, so a legitimately zero-length entry is left alone and a second run is a no-op.

**The trial-reminder monitor could only ever report OK.** `checkTrialReminders` returned `void` while the other three cron jobs return `{ degraded }`, and `scheduler/index.ts` reads `result?.degraded` to choose the Sentry check-in status. A failing reminder query logged and continued, and the monitor stayed green. It now degrades the run.

**The search route declared an argument it could not accept.** `workspaceId` was `v.optional(v.string())`, then `workspaceAccess.fromQuery()` threw `400` whenever it could not resolve one — so the optional form was unreachable over HTTP and the cross-workspace path inside `globalSearch` was dead. The validator now says what the route requires. On the client, `useGlobalSearch` holds the query behind `skipToken` until a workspace is known, instead of firing a request that can only `400` and then reporting it to Sentry.

**The bearer-token shim skipped PATCH.** `api.on(["POST", "GET", "PUT", "DELETE"], "/auth/*")` omitted it, so a `PATCH` carrying a bearer token did not get the `x-api-key` rewrite.

**Two Drizzle relations were wrong.** `userTableRelations` declared `workspaces: many(workspaceTable)`, but `workspace` has no user column and there is no inverse relation — ownership is `workspace_member.role`. Any query using `with: { workspaces: true }` would throw, so it is removed. `taskReminderSentTableRelations` was exported but never registered on the schema object; it now is.

## Related Issue(s)

None — found while auditing the codebase, not reported.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test addition or update

`workspaceId` moving from optional to required on `GET /search` tightens the schema, but every request that omitted it already failed with `400`, so no working call changes behaviour.

## How Has This Been Tested?

- [x] Unit tests
- [x] Integration tests
- [ ] Manual testing

`tests/api-integration/time-entry-duration.test.ts` adds 7 cases: a closed entry records elapsed seconds, an open entry leaves duration unset, an inverted range is refused with a 400, a search with no workspace is refused, and the backfill repairs a closed entry, clears a running one, and leaves an already-correct one untouched.

The duration cases were checked against the bug: restoring the old `duration = 0` makes both fail.

`381 unit / 191 integration` pass locally (integration was 184 before this PR). Typecheck and `biome ci` are clean.

n/a

## Checklist

## Additional Notes

Two checklist items are deliberately unticked:

- **Authorship** — this branch was written by Claude Code, as recorded in the `Co-Authored-By` trailer. The box asserting the description is in the author's own words is left for the reviewer to tick after their own review.
- **Comments** — `AGENTS.md` asks for comments that explain constraints rather than narrate code, and the reasoning here is in the commit body and this description instead.

`apps/docs/openapi.json` is **not** regenerated here even though the search schema changed, so the committed spec still shows `workspaceId` as optional. #1646 originally carried a refresh and I dropped it: re-exporting produces a string `default: "20"` on `limit` where `main` has `20`, exports the billing routes with an invalid `responses` shape because their `describeRoute` declares none, and writes a `servers[0]` that depends on the exporting environment. The spec needs its own PR that fixes those first — ideally with a CI check that the committed file matches a fresh export, since nothing enforces that today.

The tabs-vs-spaces review comment is not applied: `biome.json` overrides the global tab setting with `javascript.formatter.indentStyle: "space"`, the file has no tab-indented lines, and `biome ci` passes as-is.

## Summary by CodeRabbit

* **New Features**
  * Time-entry durations are calculated automatically from start and end times.
  * Added ISO 8601 timestamp and calendar-date validation.
  * Time entries reject reversed ranges and excessively long durations.

* **Bug Fixes**
  * Reminder processing continues when individual checks fail.
  * Authentication routes support partial updates.

* **Search**
  * Workspace selection is required for global search.
  * Searches are skipped until a workspace is selected.


andrejsshell:
/review

@coderabbitai full review

@codex review

andrejsshell:
/review

@coderabbitai full review

@codex review

andrejsshell:
/review

@coderabbitai full review

@codex review

andrejsshell:
/review

@coderabbitai full review

@codex review

andrejsshell:
/review

@coderabbitai full review

@codex review

andrejsshell:
/review

@coderabbitai full review

@codex review

PR #1649, opened 2026-08-22
fix(api): make notifications and seat reconciliation survive replicas

andrejsshell:
## Description

Two places that quietly assume a single process. Both are invisible on one instance and wrong on more than one, which is the combination the Helm chart's `autoscaling` values invite.

**Notifications never crossed instances.** `broadcastToUser` wrote straight to the local `userConnections` map and never touched the broadcast adapter, so `NOTIFICATION_CREATED` only reached sockets on the instance that raised it. A user connected to instance B saw nothing for a notification raised on instance A — the bell only caught up on the next poll. Project broadcasts already went through the adapter; user broadcasts now do too, and local sockets are served from the subscription like the project path.

The Redis channel is `kaneo:ws-user:`, **not** `kaneo:ws:user:`. A Redis glob `*` spans colons, so the obvious name would also match the existing `kaneo:ws:*:broadcast` pattern and every user message would be parsed against the project schema and dropped. The project channel and pattern are untouched, so a rolling upgrade keeps delivering while both versions are live.

**Seat reconciliation ran on every replica at once.** `initializeScheduler()` starts all four crons on every process with no leader election. The two reminder jobs survive that because they claim each send against a unique constraint before doing any work, but `reconcileWorkspaceSeats` has no such claim — so N replicas each called the billing provider for the same drifted workspace on every tick. It now runs under an advisory lock and a replica that cannot take it skips.

The lock is session-scoped (`pg_try_advisory_lock` on a dedicated pool connection) rather than the `pg_advisory_xact_lock` used in `auth.ts` and project ordering, because it has to be held across outbound HTTP calls to the billing provider, which do not belong inside an open transaction. `withLeaderLock` is written to be reusable if another job needs it.

## Related Issue(s)

None — found while auditing the codebase, not reported.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Test addition or update

`BroadcastAdapter` gains `publishToUser` and `subscribeToUser`. Both implementations in the repo are updated; the type is internal to `apps/api`.

## How Has This Been Tested?

- [x] Unit tests
- [x] Integration tests
- [ ] Manual testing

`tests/api/ws/user-broadcast.test.ts` covers delivery to every socket a user has open, isolation from other users, and the single-instance path with no adapter initialised. `tests/api/ws/in-memory-broadcast-adapter.test.ts` gains a case asserting a user message reaches the user handler and not the project handler.

`tests/api-integration/leader-lock.test.ts` runs against real Postgres and covers the property that matters: while one caller holds the lock a second caller skips rather than running, the lock is released for the next run, it is released even when the job throws, and a different key is not blocked.

`385 unit / 188 integration` pass locally (`381 / 184` before this PR). Typecheck and `biome ci` are clean.

n/a

## Checklist

## Additional Notes

Two checklist items are deliberately unticked:

- **Authorship** — this branch was written by Claude Code, as recorded in the `Co-Authored-By` trailer. The box asserting the description is in the author's own words is left for the reviewer to tick after their own review.
- **Comments** — `AGENTS.md` asks for comments that explain constraints rather than narrate code, and the reasoning here is in the commit body and this description instead. The `kaneo:ws-user:` prefix is the one thing here that is genuinely non-obvious from the code, so it is spelled out above.

The multi-instance user-delivery path is exercised through the in-memory adapter, not a live two-instance Redis setup. The Redis side is the same publish/psubscribe shape as the project channel, with the pattern collision handled explicitly, but it has not been run against two processes.

`withLeaderLock` holds one pool connection for the duration of the job. The pool max is 10 and only seat reconciliation uses it, so this is not contended today — worth remembering if more jobs adopt it.

## Summary by CodeRabbit

- **New Features**
  - Added user-targeted WebSocket broadcasts through local and Redis-backed delivery.
  - Messages reach all active connections for a specific user while remaining isolated from others.
  - Added coordination so seat reconciliation runs on only one active server at a time.

- **Bug Fixes**
  - Improved local fallback delivery when cross-server publishing is unavailable.
  - Strengthened cleanup and recovery after broadcast or reconciliation failures.

- **Tests**
  - Added coverage for targeted delivery, user isolation, lock coordination, and failure recovery.

### DIFF
--- a/apps/api/src/scheduler/leader-lock.ts
+++ b/apps/api/src/scheduler/leader-lock.ts
@@ -1,0 +1,49 @@
+import { getDatabasePool } from "../database";
+
+export const SEAT_RECONCILIATION_LOCK = 1525;
+
+export async function withLeaderLock<T>(
+  lockKey: number,
+  run: () => Promise<T>,
+  whenHeldElsewhere: () => T,
+): Promise<T> {
+  const client = await getDatabasePool().connect();
+  let acquired = false;
+  let released = false;
+
+  try {
+    const result = await client.query<{ locked: boolean }>(
+      "SELECT pg_try_advisory_lock($1) AS locked",
+      [lockKey],
+    );
+    acquired = result.rows[0]?.locked === true;
+
+    if (!acquired) {
+      client.release();
+      released = true;
+      return whenHeldElsewhere();
+    }
+
+    return await run();
+  } finally {
+    let unlockError: unknown;
+
+    if (acquired) {
+      unlockError = await client
+        .query("SELECT pg_advisory_unlock($1)", [lockKey])
+        .then(() => undefined)
+        .catch((err) => err);
+
+      if (unlockError) {
+        console.error(
+          `Failed to release advisory lock ${lockKey}`,
+          unlockError,
+        );
+      }
+    }
+
+    if (!released) {
+      client.release(unlockError ? true : undefined);
+    }
+  }
+}
--- a/apps/api/src/scheduler/seat-reconciliation.ts
+++ b/apps/api/src/scheduler/seat-reconciliation.ts
@@ -3,6 +3,7 @@ import { isBillingEnabled } from "../billing/config";
 import { syncWorkspaceSeats } from "../billing/controllers/sync-seats";
 import db from "../database";
 import { workspaceBillingTable, workspaceUserTable } from "../database/schema";
+import { SEAT_RECONCILIATION_LOCK, withLeaderLock } from "./leader-lock";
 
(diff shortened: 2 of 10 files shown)


andrejsshell:
/review

@coderabbitai full review

@codex review

andrejsshell:
/review

@coderabbitai full review

@codex review

andrejsshell:
/review

@coderabbitai full review

@codex review

andrejsshell:
/review

@coderabbitai full review

@codex review

andrejsshell:
/review

@coderabbitai full review

@codex review

PR #1655, opened 2026-08-24
refactor(api): generate the OpenAPI spec from Zod instead of rewriting it

andrejsshell:
## Description

The OpenAPI spec was assembled by running Valibot through `hono-openapi`, merging in Better Auth's generated document, then rewriting the result with nine transforms chained nine deep inside `createApp`. Most of those transforms existed to paper over two things: Valibot's JSON Schema output (`v.date()` came out as `{}`, enums lost their `type`, `$ref`s grew siblings) and a 3.1-to-3.0 downgrade for the docs site.

This emits 3.1 natively from `@hono/zod-openapi` instead. Routes are declared with `createRoute` on the `apiRouter()` factory in `apps/api/src/openapi.ts`; request schemas live in each feature's `schema.ts` and response schemas in its `response.ts`, named with `.openapi("Name")` so they become reusable components.

`utils/openapi-spec.ts` drops from nine transforms to one URL helper. The only post-processing left is a single loop injecting the shared `401` that every authenticated route returns, rather than repeating it on 154 operations.

| | before | after |
|---|---|---|
| spec version | 3.0.3 (downgraded) | **3.1.0 native** |
| component schemas | 2 | **83** |
| operations missing a summary | synthesized from operationIds | **0**, all explicit |
| operationId collisions | patched at runtime | **none**, enforced by test |
| validation | Valibot + `hono-openapi` | Zod + `@hono/zod-openapi` |

Better Auth's 35 organization operations are now declared as Zod in `apps/api/src/auth-openapi.ts`, generated from its own `generateOpenAPISchema()` output and reviewed. That replaced a runtime call to that generator plus ~200 lines of operationId/summary/tag normalization and `$ref` pruning. Regenerate and diff that file when upgrading Better Auth.

Removed: `hono-openapi`, `@hono/standard-validator`, `@valibot/to-json-schema`. Valibot remains only for internal, non-HTTP config validation under `plugins/` and `ws/`, which never reaches the spec. `AGENTS.md` is updated — it still told contributors to validate with Valibot.

## Related Issue(s)

/
## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] Documentation update
- [x] Refactoring (no functional changes) — with two deliberate behavior changes, described in Additional Notes
- [ ] Performance improvement
- [ ] Test addition or update

## How Has This Been Tested?

- [x] Unit tests — 385 passing
- [x] Integration tests — six new spec tests in `tests/api-integration/openapi.test.ts`
- [ ] Manual testing
- [x] Other (please describe): `tsc --noEmit` on `apps/api` and both `apps/web` tsconfigs; `biome ci .` clean; RPC types verified with invariant type-equality assertions (with negative controls to prove the assertions bite) across every label endpoint and fetcher.

N/A — no UI changes.

## Checklist

- [ ] I have performed a self-review of my own code

The two unticked boxes are attestations only the author can make: this description was written by Claude Code, not in my own words, and the self-review is still mine to do.

### Two behavior changes to review

1. **Authorization now runs before validation.** `createRoute({ middleware })` registers middleware ahead of the request validators (`@hono/zod-openapi` `index.mjs:147`), whereas the old routes listed `validator()` first. An unauthorized caller with a malformed body now gets `403` instead of `400`, and no longer learns whether their body parsed. Inline middleware that read `c.req.valid()` was rewritten to read the raw request, since that is not populated when it runs — this affected `task-relation` and both issue-import routes.

2. **Six web fetchers now tag the response status.** Declaring error responses puts them in each route's typed-response union, so an untagged `InferResponseType<typeof client.x.$get>` widens from `Task` to `Task | string`. Those call sites now pass `, 200`, matching what `src/types/project` already did. Fetchers that narrow on `response.ok` before `.json()` were unaffected.

### Documentation bugs this surfaced

Typing the responses against the controllers exposed places where the published spec did not match the code:

- `listGitHubRepositories` was documented as returning a bare array, but has always returned `{repositories, installations, total}`. The web client already read `.repositories`, so the spec was the wrong half.
- `deleteGitHubIntegration` returns `{success, message}`, not the integration.
- `POST /notification` returns **null** when the user has muted that notification category.
- The search response described a shape `globalSearch` never produced.
- Task `startDate`/`dueDate` were declared optional; the columns are nullable-always-present.
- `GET /api/config` is mounted before the auth middleware, so it is public — it was documented as requiring a bearer token.

### Audits, and what they caught

Two audits were written to check the migration rather than trust it, and both found real problems:

- An **authorization audit** comparing every route's middleware against `HEAD` caught that gitea's `GET /project/{projectId}` uses `workspaceAccessMiddleware(...)` directly rather than the `workspaceAccess.*` helper. The extractor had missed it, and that route would have shipped with no authorization.
- A **route-path audit** against `HEAD` confirmed no route was dropped, after the test suite caught `POST /mcp/token` going missing during the mcp rewrite.

### Not done: Redocly linting

`pnpm-workspace.yaml:33` pins `js-yaml: '>=4.2.0 <5'` and `@redocly/openapi-core` 2.46+ needs js-yaml 5. That pin looks like a deliberate security constraint, so I left it alone rather than widen it workspace-wide. The structural checks that don't need Redocly are in the integration test: 3.1 validity, global bearer auth, unique operationIds, a summary on every operation, named components, and a 401 on every non-public operation. Relaxing the pin to `>=4.2.0` is the one-line change if the full lint is wanted.

## Summary by CodeRabbit

* **New Features**
  * Updated API documentation to OpenAPI 3.1 with clearer endpoint descriptions, response formats, authentication requirements, and reusable schemas.
  * Added consistent validation and documented error responses across activities, tasks, projects, comments, integrations, notifications, billing, and workspace features.
  * Improved integration responses by masking sensitive credentials and clarifying configuration status.
  * Telegram integrations now enable unspecified event notifications by default.
  * Added clearer schemas for search, invitations, avatars, time entries, and task relations.
* **Bug Fixes**
  * Improved client handling of successful API responses and task-board requests.


andrejsshell:
All eight review findings are addressed in `6bfe74de` and `a581bdd2`. CI is green, including the integration job that was failing.

**Codex — `c.req.valid()` in task permission middleware (both comments).** Correct, and it was the cause of the 14 integration failures. `createRoute({ middleware })` registers middleware ahead of the validators, so `requireBulkTaskPermission`, `requireBulkTaskEntitlement` and `requireTaskAssigneePermission` all destructured `undefined` and 500'd. They now read the raw body, and an unknown bulk `operation` is rejected with a 400 instead of silently falling through to the `task:update` branch. Reproduced the failures locally against Postgres before fixing, then confirmed 224/224 integration tests pass.

An earlier sweep for this pattern only covered `src/*/index.ts`, which is why these were missed — they live in a controller file. The sweep now covers every file, and finds nothing else.

**Qodo #6 — project entitlement check removed.** Correct, and the most serious finding here. `POST /project` lost `requireEntitlement` in the migration. Restored, after the workspace-access middleware so `workspaceId` is populated. The middleware audit that should have caught this only matched authorization helpers, not `requireEntitlement`; it now matches any `require*` / `scopeTo*` / access helper and reports nothing else lost across the 25 migrated modules.

**Qodo #1–#5 — schemas bypassing `schema.ts`.** Fair, and inconsistent with the layout documented in `AGENTS.md`. Request schemas moved into each feature's `schema.ts` for discord, slack, generic-webhook, github and gitea, and the two inline 404 bodies moved into `response.ts`.

Also in this round: billing routes are now `hide: true`, so they stay served but no longer appear in the published document — billing is Cloud-only and shouldn't be advertised in the API reference self-hosters read.

andrejsshell:
@codex review

andrejsshell:
@coderabbitai review

andrejsshell:
Thanks — most of these were right, and one of them found a flaw in how I built `auth-openapi.ts`. Addressed in `6500877b`.

**Fixed**

- **Better Auth request schemas.** Right, and the underlying cause was worse than the symptom: I generated that file from the committed `apps/docs/openapi.json`, which had drifted from the code. Regenerated from the live `generateOpenAPISchema()`. `teamId` is no longer required on `invite-member`. My emitter also had no `anyOf` case, so `role` had silently degraded to `z.unknown()` and dropped out of `required` entirely; it is now the `string | string[]` union Better Auth actually declares. All 35 operations keep their paths and synthesized operationIds — verified by diffing against the pre-PR spec.
- **task-relation 409.** Confirmed: `create-task-relation.ts:91` throws 409 for a duplicate. Was documented as 400. Also added 404 for a missing target task.
- **column 409.** Confirmed: `create-column.ts` throws 409 for reserved and duplicate slugs.
- **`page`/`limit` NaN.** Confirmed and worth fixing — `?page=abc` reached the Drizzle `limit`/`offset` clause as `NaN`. Both now reject anything that is not a positive integer, with `limit` capped at 200. This predates the PR; the Valibot version had the same gap.
- **`userName` on time entries.** Confirmed: the list route joins the user, the single-entry route does not, so only the list item schema carries it.
- **Telegram description.** Confirmed, and it applies more widely than flagged: none of the six `validate*Config` functions make a network call, so the "verified before it is stored" wording was wrong for Discord, Slack, Gitea, GitHub and the generic webhook too. Corrected in all of them.

**Not changed, with reasons**

- **`verifyRoute` has no `projectId` path parameter (Critical).** I do not think this one holds. `workspaceAccess.fromProject` does not read only the path parameter — `workspace-access-middleware.ts:73` resolves the id as `c.req.param(source.idKey) || idFromBody`, where `idFromBody` comes from the raw JSON body at line 66-68. `POST /verify` has no path param, so it falls through to `projectId` in the body and `workspaceId` is set. This is also unchanged from before the migration: the original route carried the same middleware with `projectId` in the body. Happy to be shown otherwise if you are reading a different code path.
- **`column` update 404 and `workflow-rule` delete 404.** These statuses are unreachable. Both routes are guarded by `workspaceAccess.fromColumn` / `fromWorkflowRule`, which look the resource up first and throw `400 "Workspace ID could not be determined"` when it does not exist, so the controller's 404 never fires. Documenting them would describe a response the API cannot return. The `create` 409 you flagged alongside them is reachable and was added.
- **billing 502.** The billing routes are now `hide: true` and no longer appear in the document at all — billing is Kaneo Cloud only, so advertising it in the reference self-hosters read is noise. There is no contract left to correct.
- **OAuth `idToken` exposure.** A fair concern, but it is pre-existing behavior that this PR did not introduce or change, and the web client's logout flow depends on the current response. Changing it is an API break that deserves its own PR rather than being folded into a refactor. Flagged to the maintainer for a separate decision.
- **`has-permission` XOR.** Better Auth's own generator declares `required: ["permissions"]` with both `permission` and `permissions` present. This file's contract is to mirror that output faithfully so it can be regenerated on upgrade; encoding a stricter XOR by hand would drift from the generator on the next Better Auth bump. Worth raising upstream instead.

CI is green: unit 385, integration 224, typecheck, lint, build, docker-build.

andrejsshell:
All four addressed in `16281884`. They shared one root cause worth naming: the generator behind `auth-openapi.ts` handled a narrow slice of JSON Schema and silently fell back to `z.unknown()` for anything else, so several Better Auth contracts came out weaker than they really are. Fixed the emitter rather than patching the output.

- **Nullable organization fields.** Right. Better Auth emits these as 3.1 `type: ["string", "null"]`, and my emitter stripped the null branch without adding `.nullable()`. `logo`, `organizationId` and `teamId` accept null again. `organizationSlug` stays optional and non-null.
- **`delete-role` / `update-role` bodies.** Right, and the cause was `allOf` being unhandled — both bodies are `allOf: [{organizationId}, anyOf: [{roleName}, {roleId}]]`, which collapsed to "any JSON". Now emitted as an intersection of the base object with the roleName/roleId union, so the choice is expressed rather than lost.
- **`get-invitation` `id` required.** Right, and a regression I introduced in the previous commit. Better Auth emits it optional; the pipeline this file replaces explicitly patched it to required, and regenerating dropped that. Restored as an explicit override.
- **`permission` values.** Directionally right, container not quite. Better Auth declares `create-role`'s `permission` as an object with `additionalProperties: {type: "array", items: {type: "string"}}` — a resource-to-actions map, not a bare array. My emitter had been ignoring `additionalProperties` entirely, so it was `Record<string, unknown>`; it is now `Record<string, string[]>`. It was already required, so no change there.

`z.unknown()` fallbacks in that file dropped from 30 to 16, and the remainder are fields Better Auth genuinely declares as untyped.

Verified against the live `generateOpenAPISchema()` output, not the committed spec. CI green: unit 385, integration 224, typecheck, lint, build, docker-build.

andrejsshell:
@codex review

andrejsshell:
@coderabbitai review

PR #1702, opened 2026-09-08
fix(api): skip archived tasks in due date reminders

### LINKED ISSUE #1696: [Bug]: Getting notifications for archived tasks
### Bug Description

If there is a due date on a task, notification fires even when task is already archived (and probably in done status).

### Steps To Reproduce

1. Create a task and set a due date in the future.
2. Mark task as done.
3. Archive task.
4. Wait for due date to come.

### Expected Behavior

Notifications are sent only for open tasks.

### Actual result

Notifications are sent always.

### Screenshots

_No response_

### Environment

- Kaneo v2.22.0
- Docker
- Postgresql

### Additional Context

I'm not sure if it's a bug but in my opinion this behaviour does not make sense.

mmilanovic4:
## Description

Due date reminders were still firing for archived tasks. The scheduler filters out tasks in a final column, but archiving isn't a column, so archived tasks slipped past that check. Added a `status != "archived"` condition.

The same filter was missing in `project-webhook-reminders.ts` — fixed it as well. Added integration tests covering both schedulers.

## Related Issue(s)

Fixes #1696

## Type of Change

## How Has This Been Tested?

- [x] Integration tests

## Checklist

## Additional Notes

I left `planned` out of this. Those tasks aren't archived and I think they should still get reminders, but happy to change it if you want it excluded as well.

### DIFF
--- a/apps/api/src/scheduler/due-date-reminders.ts
+++ b/apps/api/src/scheduler/due-date-reminders.ts
@@ -1,4 +1,4 @@
-import { and, between, eq, isNotNull, isNull, or, sql } from "drizzle-orm";
+import { and, between, eq, isNotNull, isNull, ne, or, sql } from "drizzle-orm";
 import db from "../database";
 import {
   columnTable,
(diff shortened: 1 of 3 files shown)


tinsever:
great! thx for your pr!

Issue #1718, opened 2026-09-11
No way to change your password from account settings


thejdubb02:
There is no screen to change your account password. The API and the web client already support it (authClient.changePassword, and there is a useChangePassword hook in apps/web), but nothing in the UI uses it, so a self-hosted user who signed up with email and password cannot rotate their password without going into the database.

I would like to add a Security tab under Account settings with a simple change-password form: current password, new password, confirm. It would only show for accounts that actually have a password. OAuth, anonymous, and passwordless (magic link or email code) sign-ins would see a short note instead, since there is nothing to change for those.

I already have this built and tested against the current main and can open the PR right away if this sounds good. Happy to adjust the placement or wording to fit how you would want it.

PR #1719, opened 2026-09-11
feat(web): add a change-password screen under account settings

thejdubb02:
Implements #1718

## Description

Adds a Security tab under Account settings with a form to change your password. The backend and the useChangePassword hook already existed, but no screen used them, so email and password users had no way to rotate their password from the UI.

The form only appears for accounts that have a password credential. OAuth, anonymous, and passwordless (magic link or email code) accounts see a short note that password change is not available for their sign-in method. That mirrors better-auth's own server-side check (it looks for an account with providerId "credential"). Fields are current, new, and confirm, with a show/hide toggle like the sign-up form. The new password must be at least 8 characters, must differ from the current one, and must match the confirmation.

## Related Issue(s)

Implements #1718

## Type of Change

## How Has This Been Tested?

- [x] Unit tests

A component test covers the credential guard (form hidden for a non-credential account), the load-error state, both validation blocks (mismatch and new-equals-current), and a valid submit. The full apps/web test suite passes, and typecheck, biome, and the production build are clean.

## Checklist

- [x] I have added tests that prove the feature works

## Additional Notes

The i18n change adds settings.securityPage strings to en-US, with the other locales backfilled to the English text by the repo's own scripts/i18n/check.mjs --fix (ready for translation). I deliberately left use-send-password-reset.ts alone; that is a separate issue.

## Summary by CodeRabbit

- **New Features**
  - Added a Security section to account settings.
  - Users can change their password with visibility controls, validation, loading states, and success or error feedback.
  - Changing a password revokes other active sessions while keeping the current session signed in.
  - Accounts without password-based sign-in see an explanatory notice.
  - Added localized Security settings content across supported languages.

- **Tests**
  - Added coverage for password-change validation, account states, loading errors, submission, and feedback.


thejdubb02:
Thanks for the review. Both substantive points are addressed in 4febc45:

- changePassword now sets revokeOtherSessions: true, so changing the password signs out the account's other sessions while keeping the current one (better-auth's documented behavior).
- The accounts query is keyed by the current user id and only enabled once it is known, so a replaced in-app session cannot reuse the previous user's credential state. This matters because the global query client has refetchOnMount off.

I left the explicit JSX.Element return types off on purpose: the codebase omits them almost everywhere (about 4 of 244 components use them), so adding them here would cut against the prevailing style. Happy to add them if you'd prefer the annotation.

tinsever:
This UI-screenshot was auto-made by a beta tool made by @tinsever

| Description | Screenshot |
| --- | --- |
| Account settings sidebar now includes a Security tab; selecting it shows a password change form. | Screenshot |
| Account Security page showing a change-password form with current, new, and confirm fields and an Update password button. | Screenshot |
| A screenshot of the 'Security' settings page with a password change form, showing mismatched input fields. | Screenshot |

